PR #15, opened 2019-01-23
[AKS] Feature/am 50/configuring testing service

zielinskikamil:
changed lib artifact id from amlib to am-lib
added migrations files + flyway
consume and configure am-lib in am-lib-testing-api

### DIFF
--- a/am-lib/build.gradle
+++ b/am-lib/build.gradle
@@ -46,7 +46,6 @@ repositories {
 dependencies {
   compile group: 'org.jdbi', name: 'jdbi3-core', version: '3.0.0-beta3'
   compile group: 'org.jdbi', name: 'jdbi3-postgres', version: '3.0.0-beta3'
-  compile group: 'org.postgresql', name: 'postgresql', version: '42.2.5'
 
   testCompile group: 'junit', name: 'junit', version: '4.12'
 
--- a/am-lib/build.gradle
+++ b/am-lib/build.gradle
@@ -2,6 +2,7 @@ plugins {
   id 'org.owasp.dependencycheck' version '3.1.2'
   id 'com.github.ben-manes.versions' version '0.17.0'
   id 'com.jfrog.bintray' version '1.8.4'
+  id 'com.github.johnrengelman.shadow' version '4.0.4'
 }
 
 group 'uk.gov.hmcts.reform'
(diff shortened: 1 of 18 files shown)


am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DummyService.java
```diff
@@ -1,8 +1,33 @@
...
+
+import java.util.List;
+
 public class DummyService {
+    private Jdbi jdbi;
+    private String msg = "Hello Dummy Service 2";
+
+    public DummyService(String dbUrl, String user, String password) {
+        if (dbUrl == null && user == null && password == null) {
+            msg = "db url or user or pass is null";
```
r-kruk-kainos: Please put || here instead of &&

am-lib/build.gradle
```diff
@@ -88,7 +96,7 @@ publishing {
       artifact sourcesJar
       artifact javadocJar
       groupId project.group
-      artifactId 'amlib'
+      artifactId 'am-lib'
```
radoslawlandowskikainos: Is this change reflected in AM-API's dependencies?

zielinskikamil: yep

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DummyService.java
```diff
@@ -1,8 +1,33 @@
...
 
+import org.jdbi.v3.core.Jdbi;
+
+import java.util.List;
+
 public class DummyService {
+    private Jdbi jdbi;
+    private String msg = "Hello Dummy Service 2";
+
+    public DummyService(String dbUrl, String user, String password) {
```
leszekgu: It would be good to add another constructor with usage of javax.sql.DataSource;
so that in the service there would be two things:
Config class with:
```    @Bean
    @ConfigurationProperties("spring.datasource")
    public DataSource dataSource() {
        return DataSourceBuilder.create().build();
    }
```

and usage as `@Autowired DataSource ds;`

`Jdbi.create(ds);` <- which btw could be put into a separate class/Bean config with PostgresPlugin attached/installed.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DummyService.java
```diff
@@ -1,8 +1,33 @@
...
+            jdbi = Jdbi.create(dbUrl, user, password);
+        }
+    }
 
     public String getHello() {
-        return "Hello Dummy Service 2";
+        StringBuffer newMsg = new StringBuffer(msg);
+
+        if (jdbi != null) {
+            List<Integer> ids = jdbi.withHandle(handle ->
```
leszekgu: just as an idea, to show that something is really happening and remove the need to manually insert data, maybe add this:
```
        jdbi.useHandle(handle -> {
           handle.execute("INSERT INTO ACCESS_CONTROL (null)");
        });
```

zielinskikamil: the code that is already there also shows that it works without inserting. 
Hym I would like to keep it as it is tbh as we dont have access to db on aat

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DummyService.java
zielinskikamil: I will use this pattern in am-api

zielinskikamil: would you like to add this code in seperate PR and bump with new version as I've already  released 0.0.5 version ?

PR #16, opened 2019-01-23
[AKS] Add perf tests to pipeline

arompa:
Works on [PREVIEW].
Won't work as part of [AKS] pipeline.

### DIFF
--- a/src/gatling/conf/recorder.conf
+++ b/src/gatling/conf/recorder.conf
@@ -1,0 +1,53 @@
+recorder {
+  core {
+    #mode = "Proxy"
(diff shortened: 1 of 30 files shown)


src/gatling/conf/recorder.conf
```diff
@@ -0,0 +1,53 @@
+recorder {
```
leszekgu: It looks like every property of this file is commented out, though the template stays. What's the point? (pure curiosity)

arompa: Just so one knows where to config things. I am happy to remove them if you'd prefer ;)

PR #20, opened 2019-01-25
[AKS] Add a hello world test

arompa:
Added a test calling the RootController to check if the app can connect with the db.


README.md
```diff
@@ -129,6 +129,8 @@ by executing the following command:
   docker-compose up
 ```
 
+// The port should be 2704, shouldn't it?
```
zielinskikamil: should be even different as 2703 and 2704 are used by api ;/

config/pmd/ruleset.xml
```diff
@@ -1,75 +1,107 @@
...
+          class="net.sourceforge.pmd.lang.rule.XPathRule">
+        <description>
+            RestAssured tests should have assertions either using 'assert...' or 'then'
+        </description>
+        <priority>3</priority>
+        <properties>
+            <property name="xpath">
+                <value>
+                    <![CDATA[
+//ClassOrInterfaceType[
```
zielinskikamil: is it needed? or can we format it as least?

config/pmd/ruleset.xml
```diff
@@ -1,75 +1,107 @@
...
+@Image='Test'
+]/../../../MethodDeclaration[
+not(descendant::*[(starts-with(@Image, 'assert') or
+starts-with(@Image, 'Assert') or starts-with(@Image, 'then'))])
+]
+]]>
+                </value>
+            </property>
+        </properties>
+        <!--<example><![CDATA[]]></example>-->
```
zielinskikamil: this line can be removed

config/pmd/ruleset.xml
arompa: Yes - this is needed. That's how the custom rule is defined

PR #24, opened 2019-01-28
[AKS] Test aks pipeline without perftest

arompa:
hmcts/cnp-jenkins-library#392


arompa:
No longer needed - hmcts/cnp-jenkins-library#392 merged

PR #30, opened 2019-01-30
[AKS] - Feature/am 9/explicit permissions

radoslawlandowskikainos:
Should we validate the int value passed into 'permissions' column?
Is it ok at this point to keep explicitPermissions as ExplicitPermissions class?

Example create resoucre access body:

{
	"resourceId": "123",
	"accessorId": "SIGNOF",
	"explicitPermissions": {"userPermissions": ["READ", "CREATE", "UPDATE"]}
}


am-lib/src/main/java/uk/gov/hmcts/reform/amlib/enums/Permissions.java
```diff
@@ -0,0 +1,33 @@
+package uk.gov.hmcts.reform.amlib.enums;
+
+import java.util.List;
+
+public enum Permissions {
+    HIDE(0),
```
zielinskikamil: shouldn't it start with 1 and following be a power of 2? (1, 2, 4, 8, 16 etc)

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/enums/Permissions.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public int getValue() {
+        return value;
+    }
+
+    public static int sumOf(List<Permissions> permissions) {
+        int sum = 0;
+
+        for(Permissions permission: permissions) {
+            sum += permission.getValue();
```
zielinskikamil: shouldn't you here use *&=* instead of *+=* ??
I'd imagine that if I would pass .Create and .Read cases, it will produce 000011. 
With current implementation when you pass .Hide and .Create it will generate 000001 - which is wrong?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
```diff
@@ -1,13 +1,18 @@
...
+    @SqlUpdate("insert into \"AccessManagement\" (\"resourceId\", \"accessorId\", permissions) values (?, ?, ?)")
+    void createAccessManagementRecord(String resourceId, String accessorId, int permissions);
 
-    @SqlQuery("select exists(select 1 from \"AccessManagement\" where  \"accessorId\"=? and \"resourceId\"=?)")
-    boolean explicitAccessExist(String accessorId, String resourceId);
+    // The 'LIMIT 1' suffix was introduced because at the current database state (V2.2) there is a technical
+    // possibility of returning multiple records with this query, but it's forbidden from business point of view.
+    @SqlQuery("select * from \"AccessManagement\" where \"accessorId\"=? and \"resourceId\"=? LIMIT 1")
+    @RegisterConstructorMapper(AccessManagement.class)
+    AccessManagement getExplicitAccess(String accessorId, String resourceId);
```
zielinskikamil: hym I like the approach that Remi've started with binding instead of counting on order of passed parameters.
Have u seen his PR? Wdut?

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/AccessManagementServiceIntegrationTest.java
```diff
@@ -21,7 +23,11 @@ public void testSetup() {
 
     @Test
     public void createQuery_whenCreatingResourceAccess_ResourceAccessAppearsInDatabase() {
-        ams.createResourceAccess(resourceId, "dsa");
+        ams.createResourceAccess(
+                resourceId,
+                "dsa",
+                new ExplicitPermissions(Permissions.CREATE, Permissions.READ, Permissions.UPDATE)
```
zielinskikamil: plz extract them to constant

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/enums/Permissions.java
radoslawlandowskikainos: I based on documentation on confluence for that, maybe Patryk will tell more about the approach.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/enums/Permissions.java
radoslawlandowskikainos: This += is a regular arythmetic operation that adds up integers, it's not a binary one. .Create + .Read = 1 + 2 = 3

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
radoslawlandowskikainos: He uses the same parameter multiple times which forces him to do so. Here the order of question marks in SQL matches the order of method parameters so I think it's fine. It simply takes less characters to write this code this way :D

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/AccessManagementServiceIntegrationTest.java
radoslawlandowskikainos: Which ones? "Permissions.CREATE, Permissions.READ, Permissions.UPDATE" - this sequence?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
zielinskikamil: fine enough

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/AccessManagementServiceIntegrationTest.java
zielinskikamil: whole ExplicitPermissions object. its used 3 times

zielinskikamil: what if an implementation of *ExplicitPermissions* change? you will have to change 3 places instead of one. It also improves readability as in current implementation it implies that its different for each test cases which is not true.

radoslawlandowskikainos: done

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -14,22 +19,39 @@ public AccessManagementService(String url, String user, String password) {
...
-        boolean hasAccess = jdbi.withExtension(AccessManagementRepository.class,
-            dao -> dao.explicitAccessExist(userId, resourceId));
+        AccessManagement explicitAccess = jdbi.withExtension(AccessManagementRepository.class,
+            dao -> dao.getExplicitAccess(userId, resourceId));
+
+        if (explicitAccess == null) {
+            return null;
+        }
+
+        boolean hasReadPermissions = (explicitAccess.getPermissions() & Permissions.READ.getValue())
```
yarpo: what about something like this:

```
can(Permissions.READ, explicitAccess.getPermissions()); // or `hasPermissionTo(...)`

boolean can(Permissions permission, int permissionSum) {
    return permissionSum & permission.getValue());
}
```

?

yarpo: This could be reused as a utils in many places.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/db/migration/V4__Add_permissions_column_to_am.sql
```diff
@@ -0,0 +1,5 @@
+alter table "AccessManagement"
+add permissions integer not null default 0;
+
+alter table "AccessManagement"
+alter column permissions drop default;
```
yarpo: why?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/enums/Permissions.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.List;
+
+public enum Permissions {
+    HIDE(0),
+    CREATE(1),
+    READ(2),
+    UPDATE(4),
+    SHARE(8),
+    DELETE(16)
+    ;
```
yarpo: Why is this in new line?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/enums/Permissions.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private int value;
+
+    Permissions(int value) {
+        this.value = value;
```
yarpo: I wonder if we shouldn't validate it. I mean value should be: <0; 31>

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/AccessManagement.java
```diff
@@ -0,0 +1,35 @@
...
+import java.beans.ConstructorProperties;
+
+public class AccessManagement {
+
+    private final int accessManagementId;
+    private final String resourceId;
+    private final String accessorId;
+    private final int permissions;
+
+    @ConstructorProperties({"accessManagementId", "resourceId", "accessorId", "permissions"})
```
yarpo: isn't it done automatically when names are the same?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/ExplicitPermissions.java
```diff
@@ -0,0 +1,22 @@
...
+    public ExplicitPermissions(List<Permissions> userPermissions) {
+        this.userPermissions = userPermissions;
+    }
+
+    public ExplicitPermissions(Permissions...userPermissions) {
+        this.userPermissions = Arrays.asList(userPermissions);
+    }
+
+    public List<Permissions> getUserPermissions() {
+        return userPermissions;
```
yarpo: do we care for immutable? Should we?

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/AccessManagementServiceIntegrationTest.java
```diff
@@ -13,15 +15,18 @@
 
     private String resourceId;
     private final JsonNode jsonObject = JsonNodeFactory.instance.objectNode();
+    private ExplicitPermissions explicitReadCreateUpdatePermissions;
 
     @Before
     public void setupTest() {
         resourceId = UUID.randomUUID().toString();
+        explicitReadCreateUpdatePermissions = new ExplicitPermissions(Permissions.CREATE,
```
yarpo: I'd reformat it:

```
explicitReadCreateUpdatePermissions = new ExplicitPermissions(
    Permissions.CREATE, Permissions.READ, Permissions.UPDATE
);
```

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/ExplicitPermissions.java
```diff
@@ -0,0 +1,22 @@
...
+import uk.gov.hmcts.reform.amlib.enums.Permissions;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class ExplicitPermissions {
+    private final List<Permissions> userPermissions;
+
+    public ExplicitPermissions(List<Permissions> userPermissions) {
+        this.userPermissions = userPermissions;
```
yarpo: Maybe `Set<Permissions>` would be better here?

Now this can be done like that:
```
new ExplicitPermission(asList(READ, READ));
```

yes, it doens't make sens, but when we sum it up we get 4 (UPDATE) instead of 2 (READ). If we have Set it wouldn't happen.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/ExplicitPermissions.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    public ExplicitPermissions(Permissions...userPermissions) {
+        this.userPermissions = Arrays.asList(userPermissions);
```
yarpo: if we decided to use Set:

```
ImmutableSet.copyOf(list);
```

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
```diff
@@ -1,13 +1,18 @@
...
 
-    @SqlUpdate("insert into \"AccessManagement\" (\"resourceId\", \"accessorId\") values (?, ?)")
-    void createAccessManagementRecord(String resourceId, String accessorId);
+    @SqlUpdate("insert into \"AccessManagement\" (\"resourceId\", \"accessorId\", permissions) values (?, ?, ?)")
+    void createAccessManagementRecord(String resourceId, String accessorId, int permissions);
 
-    @SqlQuery("select exists(select 1 from \"AccessManagement\" where  \"accessorId\"=? and \"resourceId\"=?)")
-    boolean explicitAccessExist(String accessorId, String resourceId);
+    // The 'LIMIT 1' suffix was introduced because at the current database state (V2.2) there is a technical
+    // possibility of returning multiple records with this query, but it's forbidden from business point of view.
```
yarpo: This will be fixed soon. Maybe even in this sprint. We accept it for a very short while.

tools/postman-collections/am-lib-testing.postman_collection.json
```diff
@@ -0,0 +1,38 @@
+{
+	"info": {
```
yarpo: :+1:

tools/postman-collections/am-lib-testing.postman_collection.json
```diff
@@ -0,0 +1,38 @@
...
+						"key": "Content-Type",
+						"value": "application/json"
+					}
+				],
+				"body": {
+					"mode": "raw",
+					"raw": "{\n\t\"resourceId\": \"123\",\n\t\"accessorId\": \"SIGNOF\",\n\t\"explicitPermissions\": {\"userPermissions\": [\"CREATE\", \"UPDATE\", \"HIDE\", \"DELETE\"]}\n}"
+				},
+				"url": {
+					"raw": "localhost:2703/create-resource-access",
```
yarpo: Not for this PR

but if you have spare time at the end of this sprint can you please extract "localhost" and port to vars so that we can then reuse it for different envs (and if anything needs to be changed it's much easier to maintain).

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/db/migration/V4__Add_permissions_column_to_am.sql
yarpo: I'd leave it.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/AccessManagement.java
radoslawlandowskikainos: We'd have to setup a flag for java compiler to emit field names in a particular way. That way we would also drop using @Bind annotation in *Repository methods. I don't know however the deeper implications of setting the flag so I decided to go with annotation.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
radoslawlandowskikainos: (y)

tools/postman-collections/am-lib-testing.postman_collection.json
radoslawlandowskikainos: Yep, will keep it on my todo list, probably as a separate PR though

PR #32, opened 2019-01-30
[PREVIEW] updated port from 2703 (the same as was on api) to 3703

zielinskikamil:


README.md
```diff
@@ -129,7 +129,7 @@ by executing the following command:
   docker-compose up
 ```
 
-// The port should be 2704, shouldn't it?
+// The port should be 3704, shouldn't it?
```
arompa: Why not updating the below points of the readme and deleting the comment? :)

zielinskikamil: updated ! ;)

PR #51, opened 2019-02-07
refactor: align access_management table with design [AM-128]

swalker125:
CreateResource class is now used for CreateResourceAccess method.
Postman scripts updated.


src/gatling/bodies/createResourceAccess.json
```diff
@@ -1,7 +1,11 @@
 {
   "resourceId": "${resourceId}",
   "accessorId": "${accessorId}",
-  "explicitPermissions": {
-    "userPermissions": [ "CREATE", "READ", "UPDATE", "DELETE", "SHARE", "HIDE" ]
-  }
+  "explicitPermissions": [ "CREATE", "READ", "UPDATE", "DELETE", "SHARE", "HIDE" ],
```
yarpo: In fact this is invalid. It's either HIDE or anything else. HIDE must not be there along with other permissions.

It is addressed in AM-120, AM-121. For now we can leave it as it's not used at all. In the future we should validate it.

PR #58, opened 2019-02-12
feat: add revoke explicit access method to am-lib (AM-65)

swalker125:
created new RevokeExplicitAccess class to build object required for removal of a db entry.
Added new endpoint to testing service to hit revoke access. 
Created new postman script to revoke access.


tools/postman-collections/am-lib-testing.postman_collection.json
```diff
@@ -130,6 +130,34 @@
 				}
 			},
 			"response": []
+		},
+		{
+			"name": "Revoke Resource Access",
```
yarpo: why do we use tabs not whitespaces?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -28,6 +29,16 @@ public void createResourceAccess(ExplicitAccessRecord explicitAccessRecord) {
             dao -> dao.createAccessManagementRecord(explicitAccessRecord));
     }
 
+    /**
+     * Removes explicit access to resource accordingly to record configuration.
+     *
+     * @param removeExplicitAccessRecord an object to remove a specific explicit access record.
+     */
+    public void revokeResourceAccess(RemoveExplicitAccessRecord removeExplicitAccessRecord) {
```
yarpo: Why don't we add unit tests with H2 for this method?

tools/postman-collections/am-lib-testing.postman_collection.json
```diff
@@ -130,6 +130,34 @@
...
+				},
+				"url": {
+					"raw": "localhost:3703/lib/create-resource-access",
+					"host": [
+						"localhost"
+					],
+					"port": "3703",
+					"path": [
+						"lib",
+						"create-resource-access"
```
geek-soft: Not sure what these are but revoke request should not mention create endpoint. The same in `localhost:3703/lib/create-resource-access` above.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/AccessManagementServiceIntegrationTest.java
```diff
@@ -43,10 +45,40 @@ public void createQuery_whenCreatingResourceAccess_ResourceAccessAppearsInDataba
...
+            "select count(1) from access_management where resource_id = ?")
+            .bind(0, resourceId)
+            .mapTo(int.class)
+            .findOnly();
+
+        assertThat(count).isEqualTo(1);
+    }
+
+    @Test
+    public void revokeResourceAccess_whenRevokingResourceAccess_ResourceAccessRemovedFromDatabase() {
```
geek-soft: May I suggest having test class per method? When class have more then few methods I find it difficult to find tests that target method under change despite the fact that test methods are prepended with name of tested method.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/RemoveExplicitAccessRecord.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.hmcts.reform.amlib.models;
+
+import lombok.Builder;
+import lombok.Data;
+import org.jdbi.v3.core.mapper.reflect.JdbiConstructor;
+
+@Data
+@Builder
+public class RemoveExplicitAccessRecord {
```
geek-soft: Name of the class sounds a bit like verb which is wrong.

I suggest renaming it to `ExplicitAccessMetadata` and creating subclass `ExplicitAccessRecord` out of it.


```
ExplicitAccessRecord extends ExplicitAccessMetadata {
  // permissions only
}
```

This way we can avoid duplicating bit of code.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/AccessManagementServiceIntegrationTest.java
```diff
@@ -43,10 +45,40 @@ public void createQuery_whenCreatingResourceAccess_ResourceAccessAppearsInDataba
...
+        ams.createResourceAccess(createRecord(resourceId, ACCESSOR_ID, EXPLICIT_READ_CREATE_UPDATE_PERMISSIONS));
+
+        ams.revokeResourceAccess(removeRecord(resourceId, ACCESSOR_ID));
+
+        int count = jdbi.open().createQuery(
```
geek-soft: This is duplicated. Let's avoid duplicates in tests. People often treat unit test code as second class citizen which is wrong. Due to nature of tests they are often more complex than implementation that makes them more difficult to read. For that reason, and because we often have to fix tests broken by refactoring I suggest making tests as simple as possible. Calling method like `countResourcesById` is much easier to understand than these 5 lines.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/AccessManagementServiceIntegrationTest.java
```diff
@@ -43,10 +45,40 @@ public void createQuery_whenCreatingResourceAccess_ResourceAccessAppearsInDataba
...
+        assertThat(count).isEqualTo(0);
+    }
+
+    @Test
+    public void revokeResourceAccess_whenRevokingResourceAccessThatDoesNotExist_NoErrorExpected() {
+        ams.createResourceAccess(createRecord(resourceId, ACCESSOR_ID, EXPLICIT_READ_CREATE_UPDATE_PERMISSIONS));
+
+        ams.revokeResourceAccess(removeRecord("4", ACCESSOR_ID));
+
+        int count = jdbi.open().createQuery(
```
geek-soft: Another duplicate.

tools/postman-collections/am-lib-testing.postman_collection.json
swalker125: you're right, it just hasn't added the remove resource stuff. Odd. Good spot.

tools/postman-collections/am-lib-testing.postman_collection.json
swalker125: just how postman has exported. Agreed tabs are unecessary. Reformatted

tools/postman-collections/am-lib-testing.postman_collection.json
```diff
@@ -1,135 +1,163 @@
...
+        "method": "POST",
+        "header": [
+          {
+            "key": "Content-Type",
+            "value": "application/json"
+          }
+        ],
+        "body": {
+          "mode": "raw",
+          "raw": "{\n    \"resourceId\": \"my-super-resource-id3\",\n    \"accessorId\": \"UNIQUE-USER-ID\",\n    \"explicitPermissions\": [\"CREATE\", \"READ\", \"UPDATE\"],\n    \"accessType\": \"user\",\n    \"serviceName\": \"Service 1\",\n    \"resourceType\": \"Resource Type 1\",\n    \"resourceName\": \"resource\",\n    \"attribute\": \"\",\n    \"securityClassification\": \"Public\"\n}"
```
geek-soft: Please revert resource ID change.

PR #61, opened 2019-02-18
feat: allow granting access in transaction to more then one attribute at once (AM-99)

swalker125:


am-lib/src/test/java/uk/gov/hmcts/reform/amlib/helpers/TestConstants.java
```diff
@@ -23,10 +26,25 @@
     public static final String OTHER_ACCESSOR_ID = "b";
     public static final Set<Permission> EXPLICIT_READ_CREATE_UPDATE_PERMISSIONS =
         Stream.of(CREATE, READ, UPDATE).collect(toSet());
+    public static final Set<Permission> EXPLICIT_READ_PERMISSION = Stream.of(READ).collect(toSet());
+    public static final Set<Permission> EXPLICIT_CREATE_PERMISSION = Stream.of(CREATE).collect(toSet());
+    public static final Map<JsonPointer, Set<Permission>> EMPTY_ATTRIBUTE_PERMISSIONS = new ConcurrentHashMap<>();
+    public static final Map<JsonPointer, Set<Permission>> SINGLE_ATTRIBUTE_PERMISSION = new ConcurrentHashMap<>();
+
+    static {
```
geek-soft: Instead of this I would prefer to have some collection factory that simplifies processionals of instantiating lists, sets and maps. Something modelled after Guava factories.

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/helpers/TestConstants.java
```diff
@@ -23,10 +26,25 @@
     public static final String OTHER_ACCESSOR_ID = "b";
     public static final Set<Permission> EXPLICIT_READ_CREATE_UPDATE_PERMISSIONS =
         Stream.of(CREATE, READ, UPDATE).collect(toSet());
+    public static final Set<Permission> EXPLICIT_READ_PERMISSION = Stream.of(READ).collect(toSet());
+    public static final Set<Permission> EXPLICIT_CREATE_PERMISSION = Stream.of(CREATE).collect(toSet());
+    public static final Map<JsonPointer, Set<Permission>> EMPTY_ATTRIBUTE_PERMISSIONS = new ConcurrentHashMap<>();
+    public static final Map<JsonPointer, Set<Permission>> SINGLE_ATTRIBUTE_PERMISSION = new ConcurrentHashMap<>();
```
geek-soft: I do not like names like `SINGLE_ATTRIBUTE_PERMISSION` or `MULTIPLE_ATTRIBUTE_PERMISSIONS`. When I was reading tests code it was difficult to understand change without knowing what `SINGLE_ATTRIBUTE_PERMISSION` and `MULTIPLE_ATTRIBUTE_PERMISSIONS` do.

Maybe we should avoid using constants for it and every test should explicitly create new instance. For example instead of

```
ams.grantExplicitResourceAccess(createRecord(resourceId, ACCESSOR_ID, MULTIPLE_ATTRIBUTE_PERMISSIONS));
```

we could do:

```
... permissions = MapBuilder.
  .put("", EXPLICIT_READ_CREATE_UPDATE_PERMISSIONS)
  .put("/name", EXPLICIT_READ_PERMISSION)
  .build()

ams.grantExplicitResourceAccess(createRecord(resourceId, ACCESSOR_ID, permissions));
```

Constants hide too much IMO. They ate fine for things that does not matter like security classification. Permissions and attributes that are at core should be more visible.

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/helpers/TestDataFactory.java
```diff
@@ -1,42 +1,50 @@
...
 
     private TestDataFactory() {
         //NO-OP
     }
 
-    public static ExplicitAccessRecord createRecord(String resourceId,
-                                                    String accessorId,
-                                                    Set<Permission> explicitPermissions) {
-        return ExplicitAccessRecord.builder()
+    public static ExplicitAccessGrant createRecord(String resourceId,
```
geek-soft: I would rename that factory method. Record in my mind set should only be used for DB records mapped to `ExplicitAccessRecord`.

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/helpers/TestDataFactory.java
```diff
@@ -1,42 +1,50 @@
...
-            .serviceName(TestConstants.SERVICE_NAME)
-            .resourceType(TestConstants.RESOURCE_TYPE)
-            .resourceName(TestConstants.RESOURCE_NAME)
-            .attribute("")
+            .accessorId(ACCESSOR_ID)
+            .accessType(ACCESS_TYPE)
+            .serviceName(SERVICE_NAME)
+            .resourceType(RESOURCE_TYPE)
+            .resourceName(RESOURCE_NAME)
+            .attribute("/")
```
geek-soft: I guess we can keep `""`. Sadly `/` is a bit wired construct in JSON Pointer spec. It points to `""` attribute that is almost never used.

src/main/java/uk/gov/hmcts/reform/amapi/controllers/AmLibProxyController.java
```diff
@@ -34,19 +36,27 @@
     @SuppressWarnings("unchecked") // supressing compiler warning about casting from Object to List<String>
     @PostMapping("/create-resource-access")
     public void createResourceAccess(@RequestBody Map<String, Object> amData) {
```
geek-soft: Try changing type from `Map<String, Object>` to `ExplicitAccessGrant`

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/ExplicitAccessGrant.java
```diff
@@ -0,0 +1,25 @@
...
+public class ExplicitAccessGrant {
+
+    private final String resourceId;
+    private final String accessorId;
+    private final String accessType;
+    private final String serviceName;
+    private final String resourceType;
+    private final String resourceName;
+    private final Map<JsonPointer, Set<Permission>> attributePermissions;
+    private final String securityClassification;
```
geek-soft: We should use SecurityClasification enum here to make it less error prone. Not sure if this is the right story though.

swalker125: I think there is a story for that in the backlog. It just hasn't been refined enough to play. So will leave it for now

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -79,14 +100,14 @@ public FilterResourceResponse filterResource(String userId, String resourceId, J
         }
 
         if (READ.isGranted(explicitAccess.getPermissions())) {
-            Map<String, Set<Permission>> attributePermissions = new ConcurrentHashMap<>();
-            attributePermissions.put("/", Permissions.fromSumOf(explicitAccess.getPermissions()));
+            Map<JsonPointer, Set<Permission>> attributePermissions = new ConcurrentHashMap<>();
+            attributePermissions.put(JsonPointer.valueOf("/"), Permissions.fromSumOf(explicitAccess.getPermissions()));
```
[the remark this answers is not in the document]
swalker125: Yeah I believe this will be covered in AM-12, so I will leave as is for now.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -79,14 +100,14 @@ public FilterResourceResponse filterResource(String userId, String resourceId, J
...
 
             return FilterResourceResponse.builder()
-                    .resourceId(resourceId)
-                    .data(resourceJson)
-                    .permissions(attributePermissions)
-                    .build();
+                .resourceId(resourceId)
```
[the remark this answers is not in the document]
swalker125: For sure, Becky has the same intellij code style preferences as me. Just need to make sure we are formatting before committing.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GrantAccessIntegrationTest.java
```diff
@@ -21,9 +30,43 @@ public void setupTest() {
...
+
+    @Test
+    public void grantAccess_whenCreatingResourceAccess_ResourceAccessAppearsInDatabase() {
+        ams.grantExplicitResourceAccess(createRecord(resourceId, ACCESSOR_ID, SINGLE_ATTRIBUTE_PERMISSION));
+
+        assertThat(countResourcesById(resourceId)).isEqualTo(1);
+    }
+
+    @Test
+    public void grantAccess_whenCreatingResourceAccess_EmptyAttribute() {
```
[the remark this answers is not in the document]
geek-soft: By the way we stop prefixing test method names with service method name since we separated tests into independent test classes recently. Removing `grantAccess_` for example will leave us more space for better test names.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/ExplicitAccessGrant.java
geek-soft: Ok, let's leave it for now as is.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -29,11 +32,29 @@ public AccessManagementService(String url, String user, String password) {
...
      *
-     * @param explicitAccessRecord a record that describes explicit access to resource
+     * @param explicitAccessGrant a record that describes explicit access to resource
      */
-    public void createResourceAccess(ExplicitAccessRecord explicitAccessRecord) {
-        jdbi.useExtension(AccessManagementRepository.class,
-            dao -> dao.createAccessManagementRecord(explicitAccessRecord));
+    @Transaction
+    public void grantExplicitResourceAccess(ExplicitAccessGrant explicitAccessGrant) {
+        if (explicitAccessGrant.getAttributePermissions().size() == 0) {
```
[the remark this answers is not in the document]
swalker125: just a note for myself... I need to come back and look at this.

src/main/java/uk/gov/hmcts/reform/amapi/controllers/AmLibProxyController.java
swalker125: lots less code 👍

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/ExplicitAccessGrant.java
swalker125: Also seen it is part of the next PI

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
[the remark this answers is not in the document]
swalker125: I cannot put null as the key, and the attribute cannot be null. Both through null pointer exceptions.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/FilterResourceIntegrationTest.java
```diff
@@ -31,23 +31,23 @@ void setupTest() {
     }
 
     @Test
-    void filterResource_whenRowExistWithAccessorIdAndResourceId_ReturnPassedJsonObject() {
-        ams.createResourceAccess(createRecord(resourceId, ACCESSOR_ID, EXPLICIT_READ_CREATE_UPDATE_PERMISSIONS));
+    void whenRowExistWithAccessorIdAndResourceIdReturnPassedJsonObject() throws TransactionRolledbackException {
+        Map<JsonPointer, Set<Permission>> singleAttributePermission = new ConcurrentHashMap<>();
+        singleAttributePermission.put(JsonPointer.valueOf("/"), EXPLICIT_READ_CREATE_UPDATE_PERMISSIONS);
```
geek-soft: Let's stop using "/" as test JSON path since it is almost impossible to have in real live. I am happy to have "" or "/name" for testing purposes. IMO the more real simulated scenario is the easier it is for reader to grasp what test is for.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/FilterResourceIntegrationTest.java
```diff
@@ -57,8 +57,11 @@ void filterResource_whenRowNotExistWithAccessorIdAndResourceId_ReturnNull() {
     }
 
     @Test
-    void filterResource_whenRowExistsAndDoesntHaveReadPermissions_ReturnNull() {
-        ams.createResourceAccess(createRecord(resourceId, ACCESSOR_ID, Stream.of(CREATE, UPDATE).collect(toSet())));
+    void whenRowExistsAndDoesntHaveReadPermissionsReturnNull() throws TransactionRolledbackException {
+        Map<JsonPointer, Set<Permission>> rootLevelCreatePermission = new ConcurrentHashMap<>();
+        rootLevelCreatePermission.put(JsonPointer.valueOf("/"), EXPLICIT_CREATE_PERMISSION);
```
geek-soft: Test above uses same path but variable name is different. What is more `/` does not mean root level. Let's correct name. For me `attributePermission` would be good enough.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
```diff
@@ -22,17 +29,21 @@ void setupTest() {
...
 
     @Test
-    void revokeResourceAccess_whenRevokingResourceAccess_ResourceAccessRemovedFromDatabase() {
-        ams.createResourceAccess(createRecord(resourceId, ACCESSOR_ID, EXPLICIT_READ_CREATE_UPDATE_PERMISSIONS));
+    void whenRevokingResourceAccessResourceAccessRemovedFromDatabase() throws TransactionRolledbackException {
+        Map<JsonPointer, Set<Permission>> singleAttributePermission = new ConcurrentHashMap<>();
+        singleAttributePermission.put(JsonPointer.valueOf(""), EXPLICIT_READ_CREATE_UPDATE_PERMISSIONS);
+
+
+        ams.grantExplicitResourceAccess(grantAccess(resourceId, ACCESSOR_ID, singleAttributePermission));
```
geek-soft: I am not sure refactoring of that size was needed. You could have created overloaded method `createAccessGrantForWholeDocument` which accepts only sets of permissions. Attribute "" would be hidden within factory method.  It would only require change from `createRecord` to `createAccessGrantForWholeDocument`.

This way we would avoid nastiness of:

```
Map<JsonPointer, Set<Permission>> singleAttributePermission = new ConcurrentHashMap<>();
         singleAttributePermission.put(
```

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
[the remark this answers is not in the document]
geek-soft: We should validate values of that grant object however we need to decide what tools we wanna use. Doing it manually will require lots of boilerplate code. Not a priority at the moment.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GrantAccessIntegrationTest.java
```diff
@@ -21,8 +30,42 @@ void setupTest() {
...
+
+        ams.grantExplicitResourceAccess(grantAccess(resourceId, ACCESSOR_ID, multipleAttributePermissions));
+
+        assertThat(countResourcesById(resourceId)).isEqualTo(2);
+    }
+
+    @Test
+    void whenCreatingDuplicateResourceAccessEntryIsOverwritten() throws TransactionRolledbackException {
+        Map<JsonPointer, Set<Permission>> duplicateAttributePermissions = new ConcurrentHashMap<>();
+        duplicateAttributePermissions.put(JsonPointer.valueOf("/name"), EXPLICIT_READ_CREATE_UPDATE_PERMISSIONS);
```
geek-soft: I am not sure this is good test. Are you sure DB is hit twice? Depends how `#hashCode` and `equals` are implemented on `JsonPointer`. It is likely that because maps does not hold duplicates input map has only one entry before it is mapped to record and hit DB.

Callding `ams.grantExplicitResourceAccess` twice with same instance of `ExplicitAccessGrant` will do the work.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -29,11 +33,36 @@ public AccessManagementService(String url, String user, String password) {
...
      */
-    public void createResourceAccess(ExplicitAccessRecord explicitAccessRecord) {
-        jdbi.useExtension(AccessManagementRepository.class,
-            dao -> dao.createAccessManagementRecord(explicitAccessRecord));
+    public void grantExplicitResourceAccess(ExplicitAccessGrant explicitAccessGrant)
+        throws TransactionRolledbackException {
+        if (explicitAccessGrant.getAttributePermissions().size() == 0) {
+            throw new IllegalArgumentException("Attribute permissions cannot be empty");
+        } else {
+            jdbi.useTransaction((handle) -> {
```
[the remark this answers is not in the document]
swalker125: Annotation or the method works. I am happy to just use the annotation + jdbi.useExtension

PR #73, opened 2019-02-25
feat: add default role setup API (AM-59)

swalker125:

https://tools.hmcts.net/jira/browse/AM-59

### Change description ###

WIP - new add service method.

### Checklist ###

    - feat: a new feature
    - fix: a bug fix


am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,128 @@
...
+    public void addService(@NonNull String serviceName, String serviceDescription) {
+        if (serviceName.isEmpty()) {
+            throw new IllegalArgumentException("Service name cannot be empty");
+        }
+
+        jdbi.useExtension(DefaultRoleSetupRepository.class,
+            dao -> dao.addService(serviceName, serviceDescription));
+    }
+
+    public void addRole(@NonNull String roleName, RoleType roleType, SecurityClassification securityClassification,
```
geek-soft: Can role type and security classification be null? They seem to be NOT_NULL at DB side.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,128 @@
...
+
+    public void truncateAllDefaultPermissionsForService(String serviceName, String resourceType) {
+        jdbi.useTransaction((handle) -> {
+            try {
+                DefaultRoleSetupRepository dao = handle.attach(DefaultRoleSetupRepository.class);
+                dao.deleteDefaultPermissionsForRoles(serviceName, resourceType);
+                dao.deleteResourceAttributes(serviceName, resourceType);
+            } catch (Exception e) {
+                //TODO: new exception for removing entries.
+                throw new ErrorAddingEntriesToDatabaseException(e);
```
geek-soft: Please make more generic exception. You can even use name that does not mention database in case we wanted to use something else for storage at some point.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+        });
+    }
+
+    public void truncateAllDefaultPermissionsByResourceDefinition(String serviceName,
```
geek-soft: We should validate input to every method. Maybe we can skip it on non insert methods. I am thinking to use bean validators for it in the future :)

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/enums/AccessManagementType.java
```diff
@@ -0,0 +1,6 @@
+package uk.gov.hmcts.reform.amlib.enums;
+
+public enum AccessManagementType {
+    ROLEBASED,
```
geek-soft: Please add underscore between words.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/enums/AccessManagementType.java
```diff
@@ -0,0 +1,6 @@
+package uk.gov.hmcts.reform.amlib.enums;
+
+public enum AccessManagementType {
```
geek-soft: Wouldn't `AccessType` be enough?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/enums/SecurityClassification.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.hmcts.reform.amlib.enums;
+
+public enum SecurityClassification {
+    Public,
```
geek-soft: Please use CAPITAL CASE form. Static analysis tools won't like this.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/base/IntegrationBaseTest.java
```diff
@@ -51,4 +54,43 @@ protected static int countResourcesById(String resourceId) {
...
+            .bind(0, roleName)
+            .mapTo(int.class)
+            .findOnly();
+    }
+
+    protected static int countResources(String serviceName, String resourceType, String resourceName) {
+        return jdbi.open().createQuery(
+            "select count(1) from resources where resources.service_name = ? and resources.resource_type = ? and"
+                + " resources.resource_name = ?")
+            .bind(0, serviceName).bind(1, resourceType).bind(2, resourceName)
```
geek-soft: Each bind on separate line would make it bit more readable.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/base/IntegrationBaseTest.java
```diff
@@ -8,13 +8,15 @@
...
 import uk.gov.hmcts.reform.amlib.AccessManagementService;
+import uk.gov.hmcts.reform.amlib.DefaultRoleSetupImportService;
 
 @SuppressWarnings("PMD")
 public abstract class IntegrationBaseTest {
 
     private static final PostgreSQLContainer db = new PostgreSQLContainer().withUsername("sa").withPassword("");
     private static Jdbi jdbi;
     protected AccessManagementService ams;
+    protected DefaultRoleSetupImportService defaultRoleService;
```
geek-soft: I think we should move both `AccessManagementService` and `DefaultRoleSetupImportService` to test subclasses so that each test class e.g. `FilterResourceIntegrationTest` should do something like:

```
private AccessManagementService service;

@BeforeAll
     void setup() {
         service = new AccessManagementService(
             db.getJdbcUrl(),
             db.getUsername(),
             db.getPassword()
         )
```

By the way, it makes sense to instantiate one service per test class, not one per test method as it is now. Service is stateless.

Adding more services to base test class does not make sense. It just unnecessarily adds more responsibilities to base class.

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/helpers/DefaultRoleSetupDataFactory.java
```diff
@@ -0,0 +1,42 @@
...
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.READ_PERMISSION;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_NAME;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_TYPE;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.ROLE_NAME;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.SERVICE_NAME;
+
+public class DefaultRoleSetupDataFactory {
+
+    private DefaultRoleSetupDataFactory() {
+        //NO-OP
```
geek-soft: Some people do `throw new UnsupportedOperationException("Constructing utility class is not supported");`. I think it is more obvious than using cryptic "NO-OP". Shall we start doing it?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,130 @@
...
+import uk.gov.hmcts.reform.amlib.models.DefaultPermission;
+import uk.gov.hmcts.reform.amlib.models.DefaultPermissionGrant;
+import uk.gov.hmcts.reform.amlib.models.ResourceAttribute;
+import uk.gov.hmcts.reform.amlib.repositories.DefaultRoleSetupRepository;
+import uk.gov.hmcts.reform.amlib.utils.Permissions;
+
+public class DefaultRoleSetupImportService {
+    private final Jdbi jdbi;
+
+    public DefaultRoleSetupImportService(String url, String user, String password) {
```
geek-soft: I have a suspicious that passing `DataSource` makes more sense because then library consumers can pass data source that implements connection pool.

Can you add alternate constructor that accepts solely `DataSource`?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,130 @@
...
+        this.jdbi = Jdbi.create(url, user, password);
+        this.jdbi.installPlugin(new SqlObjectPlugin());
+    }
+
+    public void addService(@NonNull String serviceName, String serviceDescription) {
```
geek-soft: All methods should be covered by Javadoc since this is public contract.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,130 @@
...
+                            .build());
+                });
+
+            } catch (Exception e) {
+                throw new ErrorAddingEntriesToDatabaseException(e);
+            }
+        });
+    }
+
+    public void truncateAllDefaultPermissionsForService(String serviceName, String resourceType) {
```
geek-soft: @swalker125 don't you think `All` is redundant? I have feeling `truncate` is enough.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
swalker125: you're right, I thought that a type check on RoleType wouldn't allow a null value through. Shall add lombok annotations

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/DefaultRoleSetupRepository.java
```diff
@@ -0,0 +1,56 @@
...
+import uk.gov.hmcts.reform.amlib.enums.RoleType;
+import uk.gov.hmcts.reform.amlib.enums.SecurityClassification;
+import uk.gov.hmcts.reform.amlib.models.DefaultPermission;
+import uk.gov.hmcts.reform.amlib.models.ResourceAttribute;
+
+@SuppressWarnings({"LineLength", "PMD"})
+public interface DefaultRoleSetupRepository {
+    @SqlUpdate("insert into services (service_name, service_description) values (:serviceName, :serviceDescription)"
+        + " on conflict on constraint services_pkey do update set service_name = :serviceName, service_description = :serviceDescription")
+    void addService(@Bind("serviceName") String serviceName, @Bind("serviceDescription") String serviceDescription);
```
geek-soft: Please remove bind, is not needed since we compile with parameter names. Binding should be derived from parameter names.

> Compiling with parameter names removes the need for the @Bind annotation. Jdbi will then bind each un-annotated parameter to the parameter’s name.

> `@SqlUpdate("insert into users (id, name) values (:id, :name)")
void insert(long id, String name);`

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
swalker125: @geek-soft can I just double check that this would be a more generic exception for removing things from the db? Any suggestions? I was thinking ErrorRemovingEntriesFromDb where the message would read something like "error, transaction has been rolled back".

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/base/IntegrationBaseTest.java
swalker125: and then each test subclass would have to define it's own instance of a PostregreSQLContainer?

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/GrantDefaultPermissionIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+class GrantDefaultPermissionIntegrationTest extends IntegrationBaseTest {
+
+    @Test
+    void whenAddResourceDefinitionIsCalledAddNewEntryIntoDatabase() {
+        defaultRoleService.addRole(
+            ROLE_NAME, RoleType.RESOURCE, SecurityClassification.Public, AccessManagementType.ROLEBASED);
+
+        defaultRoleService.grantDefaultPermission(createDefaultPermissionGrant());
+
+        assertThat(countDefaultPermissions()).isEqualTo(1);
```
geek-soft: I am afraid that checking number of rows is not sufficient. It does not give us assurance that parameters query are bound correctly.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/GrantDefaultPermissionIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static uk.gov.hmcts.reform.amlib.helpers.DefaultRoleSetupDataFactory.createDefaultPermissionGrant;
+import static uk.gov.hmcts.reform.amlib.helpers.DefaultRoleSetupDataFactory.createReadPermissionsForAttribute;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_NAME;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_TYPE;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.ROLE_NAME;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.SERVICE_NAME;
+
+class GrantDefaultPermissionIntegrationTest extends IntegrationBaseTest {
```
geek-soft: Name is misleading if we take into consideration that there are tests ensuring 2 truncate methods work.

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportServiceTest.java
```diff
@@ -0,0 +1,84 @@
...
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_NAME;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_TYPE;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.SERVICE_NAME;
+
+class DefaultRoleSetupImportServiceTest {
+    private final DefaultRoleSetupImportService defaultRoleSetupImportService = new DefaultRoleSetupImportService(
+        "", "", "");
+
+    @Test
+    void whenServiceNameNullThrowNullPointerException() {
```
geek-soft: Maybe `Name[Is]Null`?

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportServiceTest.java
```diff
@@ -0,0 +1,84 @@
...
+        assertThatExceptionOfType(NullPointerException.class)
+            .isThrownBy(() -> defaultRoleSetupImportService.addService(null, ""));
+    }
+
+    @Test
+    void whenServiceNameIsEmptyStringThrowIllegalArgumentException() {
```
geek-soft: I think saying `IsEmpty` would be enough. No need to say `String`

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportServiceTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Test
+    void whenRoleNameIsNullThrowNullPointerException() {
+        assertThatExceptionOfType(NullPointerException.class)
+            .isThrownBy(() -> defaultRoleSetupImportService.addRole(
+                null, RoleType.RESOURCE, SecurityClassification.Public, AccessManagementType.ROLEBASED));
+    }
+
+    @Test
+    void whenRoleNameIsEmptyStringThrowIllegalArgumentException() {
```
geek-soft: We should include method name in test name or split into separate classes otherwise we will have problem with duplicates.

I sort of prefer separate classes as it give us one place to rename when method name changes.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/DefaultRoleSetupRepository.java
swalker125: I believe it is needed with the on conflict on constaint section

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/AddResourceDefinitionIntegrationTest.java
```diff
@@ -0,0 +1,40 @@
...
+        defaultRoleService.addService(SERVICE_NAME, "");
+
+        defaultRoleService.addResourceDefinition(SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME);
+        defaultRoleService.addResourceDefinition(SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME);
+
+        assertThat(countResources(SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME)).isEqualTo(1);
+    }
+
+    @Test
+    void canDeleteResourceDefinitionFromTable() {
```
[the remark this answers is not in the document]
swalker125: I personally don't mind. The class name could be updated to relfect that all tests to do with resources are in one class? Or it can be all separate. What are your thoughts?

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/AddResourceDefinitionIntegrationTest.java
```diff
@@ -0,0 +1,40 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_NAME;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_TYPE;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.SERVICE_NAME;
+
+class AddResourceDefinitionIntegrationTest extends IntegrationBaseTest {
+
+    @Test
+    void whenAddResourceDefinitionIsCalledAddNewEntryIntoDatabase() {
+        defaultRoleService.addService(SERVICE_NAME, "");
```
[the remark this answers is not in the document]
swalker125: Is this maybe due to how ccd use services? Are you thinking an overloaded method where service name is just hardcoded to empty or something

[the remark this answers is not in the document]
swalker125: Should we leave this to fail in the DB or do you think I should catch and rethrow as a custom exception?

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/AddRoleIntegrationTest.java
```diff
@@ -0,0 +1,38 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.ROLE_NAME;
+
+class AddRoleIntegrationTest extends IntegrationBaseTest {
+
+    @Test
+    void whenAddRoleIsCalledAddNewEntryIntoDatabase() {
+        defaultRoleService.addRole(ROLE_NAME, RoleType.IDAM, SecurityClassification.Public,
+            AccessManagementType.ROLEBASED);
+        assertThat(countRoles(ROLE_NAME)).isEqualTo(1);
```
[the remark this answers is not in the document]
swalker125: Seems reasonable.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/AddRoleIntegrationTest.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    @Test
+    void whenCallIsMadeDuplicatingExistingRoleUpdatesExistingEntry() {
+        defaultRoleService.addRole(ROLE_NAME, RoleType.IDAM, SecurityClassification.Public,
+            AccessManagementType.ROLEBASED);
+        defaultRoleService.addRole(ROLE_NAME, RoleType.IDAM, SecurityClassification.Public,
```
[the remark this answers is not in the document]
swalker125: I can do this for role, service and permissions. Resource definition will generate a new unique value.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,211 @@
...
+        this.jdbi.installPlugin(new SqlObjectPlugin());
+    }
+
+    public DefaultRoleSetupImportService(DataSource dataSource) {
+        this.jdbi = Jdbi.create(dataSource);
+        this.jdbi.installPlugin(new SqlObjectPlugin());
+    }
+
+    /**
+     * A method to add a new service.
```
geek-soft: Usually javadoc does not start this way. No point saying `A method to`. Have a look at examples from the popular libraries e.g. Guava or Apache Commons.

Example:
> Ensures the truth of an expression involving one or more parameters to the calling method.

Also describe behaviour like an update if exists or transaction boundaries. Let's give people proper documentation please.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/exceptions/ErrorAddingEntriesException.java
```diff
@@ -1,9 +1,9 @@
 package uk.gov.hmcts.reform.amlib.exceptions;
 
-public class ErrorAddingEntriesToDatabaseException extends RuntimeException {
+public class ErrorAddingEntriesException extends RuntimeException {
     private static final long serialVersionUID = 1L;
 
-    public ErrorAddingEntriesToDatabaseException(Throwable throwable) {
+    public ErrorAddingEntriesException(Throwable throwable) {
```
geek-soft: What about having one exaction class that covers all exceptions around data persistence e.g. `PersistenceException`.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/DefaultPermissionGrant.java
```diff
@@ -0,0 +1,24 @@
...
+import uk.gov.hmcts.reform.amlib.enums.Permission;
+import uk.gov.hmcts.reform.amlib.enums.SecurityClassification;
+
+import java.util.Map;
+import java.util.Set;
+
+@Data
+@Builder
+@AllArgsConstructor
+public class DefaultPermissionGrant {
```
geek-soft: We should separate database model from rest. Sounds like separate PR.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/DefaultPermissionGrant.java
```diff
@@ -0,0 +1,24 @@
+package uk.gov.hmcts.reform.amlib.models;
+
+import com.fasterxml.jackson.core.JsonPointer;
+import javafx.util.Pair;
```
geek-soft: Please use different class. Implement your own when needed. We should not relay on JavaFX

geek-soft: Project does not even compile on newer JDK. JavaFX got removed from JDK in later versions.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
yarpo: The thing is that when you pass an object to constructor user of lib must also import this class. 

If you pass strings it's more independent

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,214 @@
...
+        this.jdbi = Jdbi.create(dataSource);
+        this.jdbi.installPlugin(new SqlObjectPlugin());
+    }
+
+    /**
+     * Creates a new unique service or updates if already exists.
+     *
+     * @param serviceName the name of the service.
+     */
+    public void addService(@NonNull String serviceName) {
```
yarpo: I'd actually provide only method with 2 args, and `description` would be nullable (and "emptable")

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,214 @@
...
+     * <p>Operation uses a transaction and will rollback if any errors are encountered whilst adding entries.
+     *
+     * @param serviceName  the name of the service to delete default permissions for.
+     * @param resourceType the type of resource to delete default permissions for.
+     */
+    public void truncateDefaultPermissionsForService(String serviceName, String resourceType) {
+        jdbi.useTransaction((handle) -> {
+            try {
+                DefaultRoleSetupRepository dao = handle.attach(DefaultRoleSetupRepository.class);
+                dao.deleteDefaultPermissionsForRoles(serviceName, resourceType);
```
yarpo: this should be more sophisticated. Let's talk :)

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,214 @@
...
+     * @param resourceName the name of the resource to delete default permissions for.
+     */
+    public void truncateDefaultPermissionsByResourceDefinition(String serviceName,
+                                                               String resourceType,
+                                                               String resourceName) {
+        jdbi.useTransaction((handle) -> {
+            try {
+                DefaultRoleSetupRepository dao = handle.attach(DefaultRoleSetupRepository.class);
+                dao.deleteDefaultPermissionsForRoles(serviceName, resourceType, resourceName);
+                dao.deleteResourceAttributes(serviceName, resourceType, resourceName);
```
yarpo: as above

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,216 @@
...
+                        .securityClassification(permissionAndClassification.getValue())
+                        .build()
+                    );
+
+                    dao.grantDefaultPermission(
+                        DefaultPermission.builder()
+                            .serviceName(defaultPermissionGrant.getServiceName())
+                            .resourceType(defaultPermissionGrant.getResourceType())
+                            .resourceName(defaultPermissionGrant.getResourceName())
+                            .attribute(attribute.toString())
```
yarpo: I wonder if we shouldn't pass `attribute` and internally call .toString()?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,216 @@
...
+                            .roleName(defaultPermissionGrant.getRoleName())
+                            .permissions(Permissions.sumOf(permissionAndClassification.getKey()))
```
yarpo: same here - pass enum, stringify internally

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,216 @@
...
+                        .build()
+                    );
+
+                    dao.grantDefaultPermission(
+                        DefaultPermission.builder()
+                            .serviceName(defaultPermissionGrant.getServiceName())
+                            .resourceType(defaultPermissionGrant.getResourceType())
+                            .resourceName(defaultPermissionGrant.getResourceName())
+                            .attribute(attribute.toString())
+                            .roleName(defaultPermissionGrant.getRoleName())
```
yarpo: as above (and below) :)

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
geek-soft: `DataSource` is one of the standard JDBC classes so import is not a problem. Using data source massively change JDBI works as it does not need to make new connection to DB for every query. When data source is used (especially one backed by connection poll) it largely improves performance and removes problem of running out of DB connections.

geek-soft: @swalker125 I am bit worried that exposing string based constructor will be error prone. I suggest either removing or maybe describing consequences in Javadoc.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
geek-soft: Any reason @yarpo to not use method overload here?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,216 @@
...
+     * Creates a new unique service or updates if already exists.
+     *
+     * @param serviceName the name of the service.
+     */
+    public void addService(@NonNull String serviceName) {
+        if (serviceName.isEmpty()) {
+            throw new IllegalArgumentException(emptyServiceError);
+        }
+
+        jdbi.useExtension(DefaultRoleSetupRepository.class,
```
geek-soft: @swalker125 instead of implementing it again that brings duplication please delegate to method with 2 args. Simply `addService(serviceName, null)` will do.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,216 @@
...
+
+    /**
+     * Creates a new unique service, with a description, or updates if already exists.
+     *
+     * @param serviceName        the name of the service.
+     * @param serviceDescription a description of the service.
+     */
+    public void addService(@NonNull String serviceName, String serviceDescription) {
+        if (serviceName.isEmpty()) {
+            throw new IllegalArgumentException(emptyServiceError);
```
geek-soft: When we remove duplicate we will be able to remove `emptyServiceError` constant as well.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,216 @@
...
+        jdbi.useExtension(DefaultRoleSetupRepository.class, dao ->
+            dao.addResourceDefinition(serviceName, resourceType, resourceName));
+
+    }
+
+    /**
+     * Creates a new resource attribute with default permissions for a role or updates if already exists.
+     *
+     * <p>Operation uses a transaction and will rollback if any errors are encountered whilst adding entries.
+     * {@link PersistenceException}
```
geek-soft: What a surprise ;) When I look at that Javadoc I cannot resist thinking that link does not fit there. There is no place for it at the moment. 

What about:

```
...
@throws PersistenceException if any errors were encountered causing transaction rollback
```

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,216 @@
...
+     *
+     * @param defaultPermissionGrant a container for granting default permissions.
```
geek-soft: I do not think dot at the end is needed :P

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,216 @@
...
+            }
+        });
+    }
+
+    /**
+     * Deletes all default permissions within a service for a given resource type.
+     *
+     * <p>Operation uses a transaction and will rollback if any errors are encountered whilst adding entries.
+     *
+     * @param serviceName  the name of the service to delete default permissions for.
```
geek-soft: Dot's here as well. They are OK in the first / main section but not when used with @param or @returns or @throws.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/DefaultRoleSetupImportService.java
```diff
@@ -0,0 +1,216 @@
...
+                dao.deleteDefaultPermissionsForRoles(serviceName, resourceType, resourceName);
+                dao.deleteResourceAttributes(serviceName, resourceType, resourceName);
+            } catch (Exception e) {
+                throw new PersistenceException(e);
+            }
+        });
+    }
+
+    /**
+     * Deletes a resource attribute.
```
geek-soft: Didn't you mean `resource definition`?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/utils/PairEntry.java
```diff
@@ -0,0 +1,32 @@
+package uk.gov.hmcts.reform.amlib.utils;
+
+import java.util.Map;
+
+@SuppressWarnings("PMD")
+//complains about setValue having a return. It is implemented this way in Map.Entry
+public class PairEntry<K, V> implements Map.Entry<K, V> {
```
geek-soft: Maybe just `Pair`?

am-lib/src/main/resources/db/migration/V9__Add_Access_Management_Type_To_Roles_Table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TYPE ACCESSMANAGEMENTTYPE AS enum ('ROLE_BASED', 'EXPLICIT');
```
geek-soft: Can we please use underscores to separate words e.g. `ACCESS_MANAGEMENT_TYPE`. We could also rename that one to be `ACCESS_TYPE` to be in sync with Java code base.

I aim to rename `SECURITYCLASSIFICATION` into `SECURITY_CLASSIFICATION` at some point as well ;)

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/base/IntegrationBaseTest.java
geek-soft: No, container can stay in `IntegrationBaseTest`. The intention of `IntegrationBaseTest` is to provide as access to DB. It should not really give us references to AM services.

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/AddServiceDefaultRolesTest.java
```diff
@@ -0,0 +1,29 @@
+package uk.gov.hmcts.reform.amlib;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+
+class AddServiceDefaultRolesTest {
+    private final DefaultRoleSetupImportService defaultRoleSetupImportService = new DefaultRoleSetupImportService(
```
geek-soft: I like short names that are unique within the context. I think simple name as `service` would do well here.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/DefaultPermissionIntegrationTest.java
```diff
@@ -0,0 +1,107 @@
...
+    void shouldNotBeAbleToCreateDefaultPermissionWhenRoleDoesNotExist() {
+        defaultRoleService.deleteRole(ROLE_NAME);
+
+        assertThatExceptionOfType(PersistenceException.class)
+            .isThrownBy(() -> defaultRoleService.grantDefaultPermission(createDefaultPermissionGrant(READ_PERMISSION)))
+            .withMessageContaining("(role_name)=(Role Name) is not present in table \"roles\"");
+    }
+
+    @Test
+    void whenAddResourceDefinitionIsCalledAddNewEntryIntoDatabase() {
```
geek-soft: Ideally we should try to use same format for test names. I like "shouldDoXWhenY" format for example.

I won't be pushing it this time thought ;)

geek-soft: The name does not look right. Method tests `grantDefaultPermission` but it says `whenAddResourceDefinition`.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/DefaultPermissionIntegrationTest.java
```diff
@@ -0,0 +1,107 @@
...
+            SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME, ATTRIBUTE, ROLE_NAME, Permissions.sumOf(READ_PERMISSION)))
+            .isEqualTo(1);
+
+        assertThat(countResourceAttributes(
+            SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME, ATTRIBUTE, SecurityClassification.PUBLIC))
+            .isEqualTo(1);
+    }
+
+    @Test
+    void whenAddResourceDefinitionIsCalledTwiceWithSameParamsOverwriteExistingRecord() {
```
geek-soft: Same here, method name mismatch tested scenario.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/ResourceDefinitionIntegrationTest.java
```diff
@@ -0,0 +1,47 @@
...
+        defaultRoleService.addResourceDefinition(SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME);
+
+        assertThat(countResources(SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME)).isEqualTo(1);
+    }
+
+    @Test
+    void whenCallIsMadeDuplicatingExistingResourceUpdatesExistingEntry() {
+        defaultRoleService.addService(SERVICE_NAME, "");
+        defaultRoleService.addResourceDefinition(SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME);
+        defaultRoleService.addResourceDefinition(SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME);
```
geek-soft: Please use different values in second call to make sure they are updated.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/RoleIntegrationTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+        assertThat(countRoles(ROLE_NAME)).isEqualTo(1);
+    }
+
+    @Test
+    void whenCallIsMadeDuplicatingExistingRoleUpdatesExistingEntry() {
+        defaultRoleService.addRole(ROLE_NAME, RoleType.IDAM, SecurityClassification.PUBLIC, AccessType.ROLE_BASED);
+        defaultRoleService.addRole(ROLE_NAME, RoleType.RESOURCE, SecurityClassification.PRIVATE, AccessType.EXPLICIT);
+
+        assertThat(countRoles(ROLE_NAME)).isEqualTo(1);
```
geek-soft: Please assert an update happened.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/ServiceIntegrationTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+        assertThat(countServices(SERVICE_NAME)).isEqualTo(1);
+    }
+
+    @Test
+    void whenDuplicateServiceUpdatesExistingEntry() {
+        defaultRoleService.addService(SERVICE_NAME, "");
+        defaultRoleService.addService(SERVICE_NAME, "Different description");
+
+        assertThat(countServices(SERVICE_NAME)).isEqualTo(1);
```
geek-soft: Please assert that service description got updated.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/utils/PairEntry.java
swalker125: PMD complaints. I guess this is caught under supress warnings now. Shall refactor and add into comments

am-lib/src/main/resources/db/migration/V9__Add_Access_Management_Type_To_Roles_Table.sql
swalker125: You're in luck. SECURITYCLASSIFICATION is now SECURITY_CLASSIFICATION

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/defaultrolesetup/ResourceDefinitionIntegrationTest.java
swalker125: This will make a new resource. It is impossible to update an existing entry with new values.

PR #80, opened 2019-03-06
feat: alter revoking permissions to cascade on given attribute (AM-147)

rebeccabrennan:
https://tools.hmcts.net/jira/browse/AM-147

### Change description ###
Updating Access Management Repository to cascade delete attributes.
Additional testing to check attributes and child attributes are deleted accordingly.


am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
```diff
@@ -35,6 +44,83 @@ void whenRevokingResourceAccessResourceAccessRemovedFromDatabase() {
...
+            .accessorId(ACCESSOR_ID)
+            .accessType(ACCESS_TYPE)
+            .serviceName(SERVICE_NAME)
+            .resourceType(RESOURCE_TYPE)
+            .resourceName(RESOURCE_NAME)
+            .attribute("/test/childTest")
+            .securityClassification(SECURITY_CLASSIFICATION)
+            .build());
+
+        assertThat(countResourcesById(resourceId)).isEqualTo(0);
```
geek-soft: That test shows it does not really work. When access is granted to the whole document and only certain attribute access is revoked it an access to the whole doc should remain.

Your logic simply removes everything from DB at any time.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
```diff
@@ -35,6 +44,83 @@ void whenRevokingResourceAccessResourceAccessRemovedFromDatabase() {
...
+            .accessorId(ACCESSOR_ID)
+            .accessType(ACCESS_TYPE)
+            .serviceName(SERVICE_NAME)
+            .resourceType(RESOURCE_TYPE)
+            .resourceName(RESOURCE_NAME)
+            .attribute("/")
+            .securityClassification(SECURITY_CLASSIFICATION)
+            .build());
+
+        assertThat(countResourcesById(resourceId)).isEqualTo(0);
```
geek-soft: Please add new test to grant access to Resource 1 and Resource 2 (two unique resources) and make sure that when you revoke access to let's say Resource 1 then rules for Resource 2 remain.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
```diff
@@ -27,7 +27,7 @@
         + "and access_management.service_name = :serviceName "
         + "and access_management.resource_type = :resourceType "
         + "and access_management.resource_name = :resourceName "
-        + "and access_management.attribute = :attribute")
+        + "and access_management.attribute like concat(attribute, '%')")
```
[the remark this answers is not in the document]
geek-soft: Confirmed, it always removes all rules for resource which is bad.

For example if you have:

```
[
"",
"/name",
"/address/city"
]
```

making call to remove `/xxx` will remove all above rules 😮

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
```diff
@@ -35,6 +44,83 @@ void whenRevokingResourceAccessResourceAccessRemovedFromDatabase() {
         assertThat(countResourcesById(resourceId)).isEqualTo(0);
     }
 
+    @Test
+    void whenRevokingResourceAccessOnRootResourceAccessRemovedFromDatabase() {
+        ams.grantExplicitResourceAccess(createGrantForWholeDocument(resourceId, READ_PERMISSION));
+        ams.revokeResourceAccess(ExplicitAccessMetadata.builder()
+            .resourceId(resourceId)
+            .accessorId(ACCESSOR_ID)
```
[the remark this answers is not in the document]
rebeccabrennan: 👍

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
rebeccabrennan: 👍

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
rebeccabrennan: 👍 tested and works for multiple scenarios.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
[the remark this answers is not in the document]
rebeccabrennan: it does 😄

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/ExplicitAccessMetadata.java
```diff
@@ -20,7 +20,7 @@ private ExplicitAccessMetadata(String resourceId,
                                    String resourceName,
                                    JsonPointer attribute,
                                    String securityClassification) {
-        super(resourceId, accessorId, accessType, serviceName, resourceType, resourceName, attribute,
+        super(resourceId, accessorId, accessType, serviceName, resourceType, resourceName, attribute + "%",
```
geek-soft: That is wrong. That class should be reusable in other places not only for revoking access.

Other thing is that when `/amount` is revoked then `/amountInPounds` for example should remained untouched. Simple ends width is not an answer. Revoking `/amount` should revoke `/amount`, `/amount/lastUpdated` etc but not `/amountInPounds` as it is absolutely different properly.

rebeccabrennan: I have updated the sql now and removed from the class and it works. 

On your second point I question how this would be done. In other instance we assume the person will be smart in what they are asking for. If they want to delete test and everything in test they would remove test/. If we were to remove the backslash at the end for a check it would remove everything and so we would not want to do this. If they want to remove test and everything in it they would pass test/ and this works.

rebeccabrennan: In my personal opinion I feel that we should have a delete and a cascade delete method separate. I feel that if someone is calling cascade delete on an attribute it is a big deal and so I feel that going forward these should be separately implemented. Documentation will need to be strong also with this current implementation as would not want someone accidentally deleting everything when they only wanted to possibly delete one attribute and accidentally deleted it and everything contained within it.

geek-soft: Requirement is simple @rebeccabrennan . When somebody deletes attribute and that attribute points to complex object then system must delete that attribute and all children of that attribute.

IMO this is reasonable behaviour. It is like cutting of branch of the tree. When you do it you do not want to remove wooden part only and leave leaves that were attached to the branch. You expect expect whole thing to go away.

rebeccabrennan: 👍

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
```diff
@@ -27,18 +39,109 @@ void setupTest() {
...
+
+    @Test
+    void whenRevokingAccessOnAttributeShouldRemoveOnlySpecifiedAttributeAndChildren() {
+        grantExplicitResourceAccess(resourceId, "/amount");
+        grantExplicitResourceAccess(resourceId, "/amount/lastUpdated");
+        grantExplicitResourceAccess(resourceId, "/amountInPounds");
+
+        revokeResourceAccess("/amount");
+
+        assertThat(databaseHelper.countExplicitPermissions(resourceId)).isEqualTo(1);
```
swalker125: looking through this again, I would like an assertion that the attribute stored has the value /amountInPounds rather than anything else. Asserting 1 entry left probably isn't enough. I don't know 100% that this is working as intended.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
```diff
@@ -32,7 +32,8 @@
         + "and access_management.service_name = :serviceName "
         + "and access_management.resource_type = :resourceType "
         + "and access_management.resource_name = :resourceName "
-        + "and access_management.attribute = :attributeAsString")
+        + "and access_management.attribute = :attributeAsString "
+        + "or access_management.attribute ilike concat(:attributeAsString, '/','%')")
```
geek-soft: Tidy up spaces please.

geek-soft: Either use `like` or have tests for behaviour given by `ilike`.

At this point regular like would be sufficient to be consistent with rest of the library. We need to look into case matching at some point through.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
```diff
@@ -27,18 +39,109 @@ void setupTest() {
...
+    void whenRevokingResourceAccessOnRootShouldRemoveFromDatabase() {
+        ams.grantExplicitResourceAccess(createGrantForWholeDocument(resourceId, READ_PERMISSION));
+        revokeResourceAccess("");
+
+        assertThat(databaseHelper.countExplicitPermissions(resourceId)).isEqualTo(0);
+    }
+
+    @Test
+    void whenRevokingResourceAccessOnSingleNestedAttributeShouldRemoveFromDatabase() {
+        grantExplicitResourceAccess(resourceId, "/test");
```
geek-soft: To be honest I would prefer to operate on more real examples e.g.

```
{
  "claimant": {
    "name": "John",
    "age": 21,
    "address": {
      "city": "London",
      "postcode": "SE1"
    }
  },
  "defendant": {
    "name": "Marry",
    "address": {
      "city": "Swansea",
      "postcode": "SA1"
    }
  },
  "amount": 100
}
```

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
rebeccabrennan: 👍

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
```diff
@@ -1,18 +1,30 @@
...
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_NAME;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_TYPE;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.SECURITY_CLASSIFICATION;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.SERVICE_NAME;
 import static uk.gov.hmcts.reform.amlib.helpers.TestDataFactory.createGrantForWholeDocument;
 import static uk.gov.hmcts.reform.amlib.helpers.TestDataFactory.createMetadata;
+import static uk.gov.hmcts.reform.amlib.helpers.TestDataFactory.createPermissions;
 
+@SuppressWarnings("PMD")
+//AvoidDuplicateLiterals multiple occurences of same string literal needed for testing purposes.
```
geek-soft: Type in `occurences`? Can you add colon after rule name? Also it would be idea to have whitespace after //.

rebeccabrennan: 👍

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
rebeccabrennan: 👍

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
rebeccabrennan: 👍

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
```diff
@@ -32,7 +32,8 @@
         + "and access_management.service_name = :serviceName "
         + "and access_management.resource_type = :resourceType "
         + "and access_management.resource_name = :resourceName "
-        + "and access_management.attribute = :attributeAsString")
+        + "and access_management.attribute = :attributeAsString "
+        + "or access_management.attribute like concat(:attributeAsString, '/','%')")
```
geek-soft: There is still missing whitespace between args of concat.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/helpers/DatabaseHelperRepository.java
```diff
@@ -30,6 +30,10 @@
     @RegisterConstructorMapper(Role.class)
     Role getRole(String roleName);
 
+    @SqlQuery("select attribute from access_management "
+        + "where access_management.resource_id = :resourceId")
+    String getResource(String resourceId);
```
geek-soft: You are not retrieving resource here. Method name is misleading!

Why not to return `ExplicitAccessRecord`?

rebeccabrennan: Changed just before you commented this.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
rebeccabrennan: Sorry didnt realise you wanted whitespace here changed now.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/helpers/DatabaseHelperRepository.java
```diff
@@ -30,6 +31,10 @@
     @RegisterConstructorMapper(Role.class)
     Role getRole(String roleName);
 
+    @SqlQuery("select attribute from access_management "
+        + "where access_management.resource_id = :resourceId")
+    JsonPointer getAttributeForExplicitAccessRecord(String resourceId);
```
geek-soft: This is wrong. Query can result in more than one row but return type is a single JsonPointer.

Please either make query specific so that only one result can be retrieved from DB or change method signature to return a list but returning `List<JsonPointer>` might not be useful / reusable in the future.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
```diff
@@ -32,7 +32,8 @@
         + "and access_management.service_name = :serviceName "
         + "and access_management.resource_type = :resourceType "
         + "and access_management.resource_name = :resourceName "
-        + "and access_management.attribute = :attributeAsString")
+        + "and access_management.attribute = :attributeAsString "
+        + "or access_management.attribute like concat( :attributeAsString, '/','%')")
```
geek-soft: OMG, this is 3rd comment to the same line. Suggestion below so no more comments are needed.

```suggestion
        + "or access_management.attribute like concat(:attributeAsString, '/', '%')")
```

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
```diff
@@ -27,18 +39,111 @@ void setupTest() {
...
+    void whenRevokingResourceAccessShouldRemoveFromDatabase() {
         ams.grantExplicitResourceAccess(createGrantForWholeDocument(resourceId, READ_PERMISSION));
         ams.revokeResourceAccess(createMetadata(resourceId));
 
         assertThat(databaseHelper.countExplicitPermissions(resourceId)).isEqualTo(0);
     }
 
+    @Test
+    void whenRevokingResourceAccessOnRootShouldRemoveFromDatabase() {
+        ams.grantExplicitResourceAccess(createGrantForWholeDocument(resourceId, READ_PERMISSION));
```
geek-soft: Any reason you did not do `grantExplicitResourceAccess(resourceId, "");` here as well for consistency?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/repositories/AccessManagementRepository.java
rebeccabrennan: 👍

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/helpers/DatabaseHelperRepository.java
rebeccabrennan: 👍

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/RevokeAccessIntegrationTest.java
rebeccabrennan: 👍

PR #86, opened 2019-03-11
refactor: update security classification name in performance tests and Postman collections

sivas23:
### Change description ###
refactor:updated security classification in create resource access json file

### DIFF
--- a/src/gatling/bodies/createResourceAccess.json
+++ b/src/gatling/bodies/createResourceAccess.json
@@ -8,5 +8,5 @@
   "attributePermissions": {
     "": ["CREATE", "READ", "UPDATE"]
   },
-  "securityClassification": "Public"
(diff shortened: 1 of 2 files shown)


src/gatling/bodies/createResourceAccess.json
```diff
@@ -8,5 +8,5 @@
   "attributePermissions": {
     "": ["CREATE", "READ", "UPDATE"]
   },
-  "securityClassification": "Public"
+  "securityClassification": "PUBLIC"
```
geek-soft: A `Public` appears in 2 additional places in postman collections. Can you update it as well?

sivas23: Added in two places in postman collections.

PR #96, opened 2019-03-15
feat: add method to return resources with create permission (AM-221)

swalker125:

https://tools.hmcts.net/jira/browse/AM-221

### Change description ###

New method in AccessManagementService to retrieve list of resource definitions a list of roles has create permissions for.

### Checklist ###

    - feat: a new feature
    - fix: a bug fix


am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -222,6 +225,18 @@ private boolean explicitAccessType(Set<String> userRoles) {
         return roleBasedAccessRecords.stream().collect(getMapCollector());
     }
 
+    /**
+     * Retrieves a list of {@link ResourceDefinition} that user roles have create permissions for.
+     *
+     * @param userRoles a set of roles
+     * @return a list of resource definitions
+     */
+    public List<ResourceDefinition> resourceCreationAllowedList(Set<String> userRoles) {
```
geek-soft: That beautiful method deserves better name. Maybe `getResourceDefinitionsWithCreatePermission`

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/repositories/AccessManagementRepository.java
```diff
@@ -53,4 +55,8 @@
 
     @SqlQuery("select access_management_type from roles where role_name = :roleName")
     AccessType getRoleAccessType(String roleName);
+
+    @SqlQuery("select service_name, resource_type, resource_name from default_permissions_for_roles where role_name = :roleName and default_permissions_for_roles.permissions = 1")
```
geek-soft: No need to prefix column name with table name if there is no ambiguity. 

```suggestion
    @SqlQuery("select service_name, resource_type, resource_name from default_permissions_for_roles where role_name = :roleName and permissions = 1")
```

geek-soft: I haven't read the story but current implementation works only when CREATE is solely permission. Of course user can create resource when CREATE permission is mixed with other permissions .e.g. resource definition with permission equal '3' that means 'CREATE` and 'READ' should also be included in outcome.

```suggestion
    @SqlQuery("select service_name, resource_type, resource_name from default_permissions_for_roles where role_name = :roleName and permissions & 1 = 1")
```

geek-soft: Adding `distinct` clause will remove duplicates so not need to remap later.

```suggestion
    @SqlQuery("select distinct service_name, resource_type, resource_name from default_permissions_for_roles where role_name = :roleName and default_permissions_for_roles.permissions = 1")
```

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/ResourceCreationAllowedListIntegrationTest.java
```diff
@@ -0,0 +1,104 @@
...
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.CREATE_PERMISSION;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.OTHER_ROLE_NAME;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.READ_PERMISSION;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_NAME;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_TYPE;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.ROLE_NAME;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.ROOT_ATTRIBUTE;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.SERVICE_NAME;
+
+class ResourceCreationAllowedListIntegrationTest extends PreconfiguredIntegrationBaseTest {
```
geek-soft: Please add test where attribute have more than CREATE permissions together with at least one more.

Please also add test where there is more than one attribute per resource with CREATE permission. In that scenario only one record in result is expected.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -222,6 +225,18 @@ private boolean explicitAccessType(Set<String> userRoles) {
...
 
+    /**
+     * Retrieves a list of {@link ResourceDefinition} that user roles have create permissions for.
+     *
+     * @param userRoles a set of roles
+     * @return a list of resource definitions
+     */
+    public List<ResourceDefinition> getResourceDefinitionsWithCreatePermission(Set<String> userRoles) {
+        return jdbi.withExtension(AccessManagementRepository.class, dao ->
+            userRoles.stream().map(dao::resourceCreationAllowedList).flatMap(Collection::stream)
```
geek-soft: I suggest having `map` and `collect` on separate lines that IMO helps understanding transformations.

Wouldn't it be better to pass set of roles to JDBI and having one query instead of n (where n is number of roles).

Does `.toList()` protect us from duplicates e.g. in scenario where role A and role B return same resource definition. If we had one query distinct would nicely dedicate on DB level without needed to read and deserialise that much.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetResourceDefinitionsWithCreatePermissionIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+            CREATE_PERMISSION, resource2, OTHER_ROLE_NAME));
+
+        Set<String> userRoles = ImmutableSet.of(ROLE_NAME, OTHER_ROLE_NAME);
+
+        List<ResourceDefinition> result = service.getResourceDefinitionsWithCreatePermission(userRoles);
+
+        assertThat(result).containsExactly(resource, resource2);
+    }
+
+    private ResourceDefinition resourceBuilder(String resourceName) {
```
geek-soft: Method should be renamed into `buildResource`. A `resourceBuilder` suggests that resource builder is returned while in fact final resource definition is returned.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetResourceDefinitionsWithCreatePermissionIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.ROLE_NAME;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.ROOT_ATTRIBUTE;
+import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.SERVICE_NAME;
+
+class GetResourceDefinitionsWithCreatePermissionIntegrationTest extends PreconfiguredIntegrationBaseTest {
+    private static AccessManagementService service = initService(AccessManagementService.class);
+    private static DefaultRoleSetupImportService importerService = initService(DefaultRoleSetupImportService.class);
+
+    private final ResourceDefinition resource = resourceBuilder(RESOURCE_NAME);
+    private final ResourceDefinition resource2 = resourceBuilder(RESOURCE_NAME + "2");
```
geek-soft: I am not big fan of using numbers to differentiate instances. It looks ugly IMO. I'd prefer something like `otherResource`.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetResourceDefinitionsWithCreatePermissionIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+        importerService.addResourceDefinition(
+            resource2.getServiceName(), resource2.getResourceType(), resource2.getResourceName());
+    }
+
+    @Test
+    void shouldRetrieveResourceDefinitionWhenRecordExists() {
+        importerService.grantDefaultPermission(createDefaultPermissionGrant(ROOT_ATTRIBUTE, CREATE_PERMISSION));
+
+        List<ResourceDefinition> result =
+            service.getResourceDefinitionsWithCreatePermission(Collections.singleton(ROLE_NAME));
```
geek-soft: Shall we use `ImmutableSet.of()` for consistency?

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetResourceDefinitionsWithCreatePermissionIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+    private final ResourceDefinition resource = resourceBuilder(RESOURCE_NAME);
+    private final ResourceDefinition resource2 = resourceBuilder(RESOURCE_NAME + "2");
+
+    @BeforeEach
+    void setUp() {
+        importerService.addRole(ROLE_NAME, RoleType.RESOURCE, SecurityClassification.PUBLIC, AccessType.ROLE_BASED);
+        importerService.addRole(
+            OTHER_ROLE_NAME, RoleType.RESOURCE, SecurityClassification.PUBLIC, AccessType.ROLE_BASED);
+        importerService.addResourceDefinition(
```
geek-soft: I'd like to look into out methods at some point and see whether we can start using ResourceDefinition in such cases. It would be nicer to reuse already exiting `resource` instance. Not for now though.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetResourceDefinitionsWithCreatePermissionIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+        assertThat(result).containsExactly(ResourceDefinition.builder()
+            .resourceName(RESOURCE_NAME)
+            .resourceType(RESOURCE_TYPE)
+            .serviceName(SERVICE_NAME)
+            .build());
+    }
+
+    @Test
+    void shouldRetrieveResourceDefinitionWhenMultipleRecordsExistWithDifferentAttributes() {
+        importerService.grantDefaultPermission(createDefaultPermissionGrant(
```
geek-soft: A `createDefaultPermissionGrant` is bit cryptic. I know that it grants access to same resource definition as `resource` instance but I think it would be better to pass what resource definition we grant access to for readability purpose.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetResourceDefinitionsWithCreatePermissionIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+        importerService.grantDefaultPermission(createDefaultPermissionGrant(CREATE_PERMISSION, resource2, ROLE_NAME));
+
+        List<ResourceDefinition> result =
+            service.getResourceDefinitionsWithCreatePermission(Collections.singleton(ROLE_NAME));
+
+        assertThat(result).containsExactly(resource2, resource);
+    }
+
+    @Test
+    void shouldRetrieveMultipleResourceDefinitionsWhenDefaultPermissionsExistForDifferentRoles() {
```
geek-soft: Please simulate scenario where resource A has permission granted to same attribute for two different roles.

```
importerService.grantDefaultPermission(createDefaultPermissionGrant(CREATE_PERMISSION, resource, ROLE_NAME));
importerService.grantDefaultPermission(createDefaultPermissionGrant(CREATE_PERMISSION, resource, OTHER_ROLE_NAME));
```

I suspect that currently two duplicate items will be returned.

swalker125: Added test case. I believe I had already fixed this before I saw this comment.

PR #102, opened 2019-03-19
refactor: getRolePermissions implementation in filterResource method (AM-168)

swalker125:

https://tools.hmcts.net/jira/browse/AM-168

### Change description ###

Refactoring of filterResource method for role based access now uses getRolePermissions method.

Refactor getRolePermissions to take in a resource definition.

### Checklist ###

    - feat: a new feature
    - fix: a bug fix


swalker125:
BASE IS SET TO AM-60!!

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -190,29 +175,26 @@ public FilterResourceResponse filterResource(@NotBlank String userId,
...
-                                                                @NotBlank String resourceName,
+    public Map<JsonPointer, Set<Permission>> getRolePermissions(@NotNull @Valid ResourceDefinition resource,
                                                                 @NotEmpty Set<@NotBlank String> userRoles) {
         List<Map<JsonPointer, Set<Permission>>> permissionsForRoles =
             jdbi.withExtension(AccessManagementRepository.class, dao ->
                 userRoles.stream()
-                    .map(role -> dao.getRolePermissions(serviceName, resourceType, resourceName, role))
+                    .map(role -> dao.getRolePermissions(resource, role))
                     .map(roleBasedAccessRecords -> roleBasedAccessRecords.stream()
+                        .filter(record -> !explicitAccessType(record.getRoleName()))
```
geek-soft: I'd suggest filtering out explicit roles first before calling `dao.getRolePermissions` or making filtering on DB side inside `dao.getRolePermissions` so that if explicit role is passed in method returns empty list.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -142,76 +144,66 @@ public void revokeResourceAccess(@NotNull @Valid ExplicitAccessMetadata explicit
...
+                .permissions(attributePermissions)
+                .accessManagementType(accessManagementType)
+                .build())
             .build();
     }
 
-    private boolean explicitAccessType(Set<String> userRoles) {
-        return jdbi.withExtension(AccessManagementRepository.class,
-            dao -> dao.getRoleAccessType(userRoles.iterator().next())).equals(AccessType.EXPLICIT);
+    private boolean roleBasedAccessType(String userRole) {
```
yarpo: why not `isRoleBasedAccess`?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -136,83 +137,72 @@ public void revokeResourceAccess(@NotNull @Valid ExplicitAccessMetadata accessMe
...
         } else {
             attributePermissions = explicitAccess.stream().collect(getMapCollector());
+            accessManagementType = AccessType.EXPLICIT;
         }
 
         JsonNode filteredJson = filterService.filterJson(resource.getResourceJson(), attributePermissions);
 
-        return FilterResourceResponse.builder()
+        return FilteredResourceEnvelope.builder()
             .resourceId(resource.getResourceId())
```
geek-soft: I think we should rename:
- `resource.getResourceId()` into `resource.getId()`
- `resource.getType()` into `resource.getDefinition()`
- `resource.getResourceJson()` into `resource.getData()`

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -136,83 +137,72 @@ public void revokeResourceAccess(@NotNull @Valid ExplicitAccessMetadata accessMe
...
+            .resourceDefinition(resource.getType())
             .data(filteredJson)
-            .permissions(attributePermissions)
+            .access(AccessEnvelope.builder()
+                .permissions(attributePermissions)
+                .accessManagementType(accessManagementType)
```
geek-soft: We should rename `.accessManagementType` into `.accessType`. I hope it makes sense for Katie. Might be worth playing it through her tomorrow before she disappears for a week.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
geek-soft: We should optimise it a bit I suppose. At the moment for 5 roles we make 5 calls to DB which might not be optimal. Simply number of DB requests depends on number of roles. Making one call with set of roles that filters out explicit roles might be more efficient. Ideally if microbenchmark confirmed it.

geek-soft: Method should be moved under `getRolePermissions` to make reading easier as now `getRolePermissions` method is the only caller.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -232,7 +222,8 @@ private boolean explicitAccessType(Set<String> userRoles) {
      */
     @SuppressWarnings("LineLength")
     @AuditLog("returned resources that user with roles '{{userRoles}}' has create permission to: {{result}}")
-    public Set<ResourceDefinition> getResourceDefinitionsWithRootCreatePermission(@NotEmpty Set<@NotBlank String> userRoles) {
+    public Set<ResourceDefinition> getResourceDefinitionsWithRootCreatePermission(
```
geek-soft: Can we revert that change? I do not think it makes it more readable.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/repositories/AccessManagementRepository.java
```diff
@@ -46,7 +46,7 @@
     @SuppressWarnings("PMD") // UseObjectForClearerAPI: more than 3 parameters makes sense for now, subject to change
     @SqlQuery("select * from default_permissions_for_roles where service_name = :serviceName and resource_type = :resourceType and resource_name = :resourceName and role_name = :roleName")
     @RegisterConstructorMapper(RoleBasedAccessRecord.class)
-    List<RoleBasedAccessRecord> getRolePermissions(String serviceName, String resourceType, String resourceName, String roleName);
+    List<RoleBasedAccessRecord> getRolePermissions(@BindBean ResourceDefinition resource, String roleName);
```
geek-soft: We should rename variable to be `resourceDefinition`. A `resource` alone is bit confusing to me.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
geek-soft: It might also make sense to have the following as next step:

```
@Data
@Builder
public final class Resource {
    @NotBlank
    private final String id;
    @NotNull
    @Valid
    private final ResourceDefinition definition;
    @NotNull
    private final JsonNode data;
}

@Data
@Builder
public final class FilteredResourceEnvelope {
    private Resource resource;
    private AccessEnvelope access;
}
```

Reason is that since envelope contains all properties from resource it make sense to reuse resource definition.

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/AccessManagementServiceValidationTest.java
```diff
@@ -73,16 +74,15 @@ void filterResourceMethodShouldRejectInvalidArguments(String userId, Set<String>
 
     @ParameterizedTest
     @ArgumentsSource(InvalidArgumentsProvider.class)
-    void getRolePermissionsMethodShouldRejectInvalidArguments(String serviceName,
-                                                              String resourceType,
-                                                              String resourceName,
+    void getRolePermissionsMethodShouldRejectInvalidArguments(ResourceDefinition resource,
```
geek-soft: Same here, let's rename to `resourceDefinition`.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -136,83 +137,72 @@ public void revokeResourceAccess(@NotNull @Valid ExplicitAccessMetadata accessMe
...
-    public Map<JsonPointer, Set<Permission>> getRolePermissions(@NotBlank String serviceName,
-                                                                @NotBlank String resourceType,
-                                                                @NotBlank String resourceName,
+    public Map<JsonPointer, Set<Permission>> getRolePermissions(@NotNull @Valid ResourceDefinition resource,
                                                                 @NotEmpty Set<@NotBlank String> userRoles) {
         List<Map<JsonPointer, Set<Permission>>> permissionsForRoles =
             jdbi.withExtension(AccessManagementRepository.class, dao ->
                 userRoles.stream()
-                    .map(role -> dao.getRolePermissions(serviceName, resourceType, resourceName, role))
+                    .filter(this::roleBasedAccessType)
```
geek-soft: Obviously that needs to be removed from the method logic since `getRolePermissions` should not reject any access type.

As a solution I'd suggest reducing list of user roles passed to `filterResource` method to contain only roles with role based access before such list be passed into `getRolePermissions` method.

geek-soft: In case `getRolePermissions` is called by `filterResource` (both public methods) two audit logs will be generated. It is obviously unwanted behaviour that we need to find work around for. Solution would be either:
* extracting private method that is called by both `getRolePermissions` and `filterResource`
* trying to reconfigure auditing aspect to kick off only when public method wasn't called by other public method from the same service

swalker125: I was leaning towards extracting common logic to a private method called by both 👍

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -136,86 +137,84 @@ public void revokeResourceAccess(@NotNull @Valid ExplicitAccessMetadata accessMe
...
      */
-    @AuditLog("returned role access to resource defined as '{{serviceName}}|{{resourceType}}|{{resourceName}}' "
-        + "for roles '{{userRoles}}': {{result}}")
-    public Map<JsonPointer, Set<Permission>> getRolePermissions(@NotBlank String serviceName,
-                                                                @NotBlank String resourceType,
-                                                                @NotBlank String resourceName,
+    @AuditLog("returned role access to resource defined as '{{resourceDefinition.serviceName}}|"
+        + "{{resourceDefinition.resourceType}}|{{resourceDefinition.resourceName}}' for roles '{{userRoles}}': "
+        + "{{result}}")
+    public Map<JsonPointer, Set<Permission>> getRolePermissions(@NotNull @Valid ResourceDefinition resourceDefinition,
```
yarpo: Why don't we create:

```
class AccessEnvelope extends Map<JsonPointer, Set<Permission>> {}
```

just to make it easier to read?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -136,86 +137,84 @@ public void revokeResourceAccess(@NotNull @Valid ExplicitAccessMetadata accessMe
...
-    public FilterResourceResponse filterResource(@NotBlank String userId,
-                                                 @NotEmpty Set<@NotBlank String> userRoles,
-                                                 @NotNull @Valid Resource resource) {
-        if (userRoles.size() > 1) {
-            throw new IllegalArgumentException("Currently a single role only is supported. "
-                + "Future implementations will allow for multiple roles.");
-        }
-
+    @AuditLog("filtered access to resource '{{resource.id}} defined as '{{resource.definition.serviceName}}|"
+        + "{{resource.definition.resourceType}}|{{resource.definition.resourceType}} for accessor '{{userId}}' "
```
geek-soft: A `resource.definition.resourceType` is bound twice.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -136,86 +137,84 @@ public void revokeResourceAccess(@NotNull @Valid ExplicitAccessMetadata accessMe
...
     }
 
     /**
      * Retrieves a list of {@link RoleBasedAccessRecord } and returns attribute and permissions values.
      *
-     * @param serviceName  name of service
-     * @param resourceType type of resource
-     * @param resourceName name of a resource
-     * @param userRoles    A set of user roles
+     * @param resourceDefinition {@link ResourceDefinition} a unique service name, resource definition and resource name
```
geek-soft: It should be more `, resource type and` not `, resource definition and`.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -136,86 +137,84 @@ public void revokeResourceAccess(@NotNull @Valid ExplicitAccessMetadata accessMe
      * @param userId    accessor ID
      * @param userRoles accessor roles
      * @param resource  envelope {@link Resource} and corresponding metadata
-     * @return envelope {@link FilterResourceResponse} with resource ID, filtered JSON and map of permissions if access
-     *         to resource is configured, otherwise null
-     * @throws PersistenceException if any persistence errors were encountered
+     * @return envelope {@link FilteredResourceEnvelope} with resource ID, filtered JSON and map of permissions if
+     *     access to resource is configured, otherwise null.
```
geek-soft: Revert missing `@throws PersistenceException if any persistence errors were encountered`

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -136,86 +137,84 @@ public void revokeResourceAccess(@NotNull @Valid ExplicitAccessMetadata accessMe
...
 
     /**
      * Retrieves a list of {@link RoleBasedAccessRecord } and returns attribute and permissions values.
      *
-     * @param serviceName  name of service
-     * @param resourceType type of resource
-     * @param resourceName name of a resource
-     * @param userRoles    A set of user roles
+     * @param resourceDefinition {@link ResourceDefinition} a unique service name, resource definition and resource name
+     * @param userRoles          A set of user roles
```
geek-soft: Since we changing it, let's start it with lowercase `a`.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/AccessEnvelope.java
```diff
@@ -1,18 +1,17 @@
...
 import java.util.Set;
 
 @Data
 @Builder
-public final class FilterResourceResponse {
-    private String resourceId;
-    private JsonNode data;
+public final class AccessEnvelope {
     private Map<JsonPointer, Set<Permission>> permissions;
+    private AccessType accessManagementType;
```
geek-soft: Please make these fields final, otherwise `@Data` will generate setters.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/FilteredResourceEnvelope.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.hmcts.reform.amlib.models;
+
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+public final class FilteredResourceEnvelope {
+    private Resource resource;
```
geek-soft: Please make these fields final, otherwise `@Data` will generate setters.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
geek-soft: One more thing, there is no `'` before ` for accessor` which results in `defined as 'Service|Resource Type|Resource Type for` (no closing apostrophe)

PR #110, opened 2019-03-27
renaming accessType enum to accessManagementType

rebeccabrennan:

https://tools.hmcts.net/jira/browse/AM-217


geek-soft:
Why is that rename needed @rebeccabrennan? A `AccessType` name fits the options and `Management` does not add any value IMO, just  making name longer.

rebeccabrennan:
renaming as we now have AccessManagementType which is EXPLICIT or ROLE_BASED and we now have AccessType which is USER or ROLE

rebeccabrennan:
AccessType changed to AccessorType in pull request feat: support for relationship between user and resource in explicit access record #112

geek-soft:
Closing as it is not viable option.

PR #113, opened 2019-03-27
build: add performance benchmark (AM-261)

swalker125:
*JIRA:* https://tools.hmcts.net/jira/browse/AM-261


am-lib/build.gradle
```diff
@@ -76,18 +76,22 @@ dependencies {
     compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.9.8'
     compile group: 'org.hibernate.validator', name: 'hibernate-validator', version: '6.0.16.Final'
     compile group: 'org.aspectj', name: 'aspectjrt', version: '1.9.2'
+    compile group: 'com.google.guava', name: 'guava', version: '27.1-jre'
+    compile group: 'org.springframework.boot', name: 'spring-boot-starter-jdbc', version: '2.1.3.RELEASE'
```
geek-soft: As discussed these dependencies should not be in main classpath as this way they would be included in library jar.

am-lib/build.gradle
```diff
@@ -76,18 +76,22 @@ dependencies {
...
+
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter', version: '5.4.1'
     testCompile group: 'org.mockito', name: 'mockito-core', version: '2.25.1'
     testCompile group: 'org.assertj', name: 'assertj-core', version: '3.12.2'
     testCompile group: 'org.flywaydb', name: 'flyway-core', version: '5.2.4'
     testCompile group: 'org.testcontainers', name: 'postgresql', version: '1.11.1'
     testCompile group: 'com.google.guava', name: 'guava', version: '27.1-jre'
-    testCompile group: 'uk.gov.hmcts.reform', name: 'java-logging', version: '5.0.1'
-    testCompile group: 'uk.gov.hmcts.reform', name: 'java-logging-appinsights', version: '5.0.1'
+    testCompile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
```
geek-soft: Change is not needed, `java-logging` brings logback as transitive dep.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementServiceBenchmarks.java
```diff
@@ -0,0 +1,118 @@
...
+                    .put(JsonPointer.valueOf("/caseName"), ImmutableSet.of(UPDATE, CREATE, READ, DELETE))
+                    .put(JsonPointer.valueOf("/children"), ImmutableSet.of(READ))
+                    .put(JsonPointer.valueOf("/children/firstChild"), ImmutableSet.of(UPDATE))
+                    .put(JsonPointer.valueOf("/children/adoption"), ImmutableSet.of(READ, CREATE))
+                    .put(JsonPointer.valueOf("/applicant"), ImmutableSet.of(READ, CREATE))
+                    .put(JsonPointer.valueOf("/solicitor"), ImmutableSet.of(UPDATE, DELETE, CREATE))
+                    .build();
+
+            //create explicit access grant
+            service.grantExplicitResourceAccess(ExplicitAccessGrant.builder()
```
geek-soft: My understanding is that explicit grant is performed before each method call performed by benchmark. Am I right?

We need to change it to do bulk upload before benchmark start / one per benchmark and clear DB in benchmark tear down.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementServiceBenchmarks.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Benchmark
+    public void baseline() {
+        //NO-OP
+    }
+
+    @Benchmark
+    public void filterResourceBenchmark(RecordState state) {
+        state.service.filterResource(state.accessorId, ImmutableSet.of("DOES_NOT_EXIST"), state.resource);
```
geek-soft: This executes explicit access only, hoverer we would probably like to mix explicit and role based or have two benchmark in worse case scenario.

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/AccessManagementServiceBenchmarkTest.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Collection;
+
+import static org.junit.Assert.assertTrue;
+
+@SuppressWarnings("PMD.JUnitTestsShouldIncludeAssert")
+class AccessManagementServiceBenchmarkTest {
+
+    private static final double REFERENCE_SCORE = 50.000;
+
+    @Test
```
geek-soft: Running this in nightly pipeline might be a challenge as performance test step is hard wired to gatling. We would need to try work around it.

PR #142, opened 2019-04-24
feat: retrieve security classifications of attributes for a role

swalker125:

https://tools.hmcts.net/jira/browse/AM-271

### Change description ###

Refactoring of getRolePermissions method. Now only takes in a String role and returns and new RolePermissions object.

RolePermissions  = permissions, accessType, securityClassifications


am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -263,6 +246,51 @@ public FilteredResourceEnvelope filterResource(@NotBlank String userId,
...
+     *
+     * @param resourceDefinition {@link ResourceDefinition} a unique service name, resource type and resource name
+     * @param roleName           user role name
+     * @return {@link RolePermissions} a combination of permissions and security classifications for a role name
+     * @throws PersistenceException if any persistence errors were encountered
+     */
+    @AuditLog("returned role access to resource defined as '{{resourceDefinition.serviceName}}|"
+        + "{{resourceDefinition.resourceType}}|{{resourceDefinition.resourceName}}' for role '{{roleName}}': "
+        + "{{result}}")
+    public RolePermissions getRolePermissions(@NotNull @Valid ResourceDefinition resourceDefinition,
```
geek-soft: I'd consider the following steps:

1. get access type and security classification for role (*)
2. calculate visible security classifications using utility
3. fetch from DB all attributes that are visible for role alongside their security classifications and permissions (*)

(*) is used to mark DB calls

Obviously there is no data structure for an output needed in step 3 but so called query class could be created in designated package.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -263,6 +246,46 @@ public FilteredResourceEnvelope filterResource(@NotBlank String userId,
...
+        Map<AccessType, SecurityClassification> roleData = jdbi.withExtension(AccessManagementRepository.class, dao ->
+            dao.getRoles(Collections.singleton(roleName), Stream.of(EXPLICIT, ROLE_BASED).collect(toSet())))
+            .stream().collect(toMap(Role::getAccessType, Role::getSecurityClassification));
+
+        if (roleData.isEmpty()) {
+            return null;
+        }
+
+        List<AttributeData> attributeData = jdbi.withExtension(AccessManagementRepository.class, dao ->
+            dao.getAttributeSecurityClassificationsForResource(
```
geek-soft: I do not feel super comfortable with `getAttributeSecurityClassificationsForResource` as it does not return security classification. Maybe `getDefaultAtrributeDefinitions` or `getDefaultAtrributePermissionsAndSecurityClassification`

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -263,6 +246,46 @@ public FilteredResourceEnvelope filterResource(@NotBlank String userId,
...
+                resourceDefinition, roleName, SecurityClassifications.fromValueOf(
+                    roleData.entrySet().iterator().next().getValue().getHierarchy())));
+
+        if (attributeData.isEmpty()) {
+            return null;
+        }
+
+        return RolePermissions.builder()
+            .accessType(roleData.entrySet().iterator().next().getKey())
```
geek-soft: A sequence of `roleData.entrySet().iterator().next()` appears quite often, shall we extract both role's access type and default security classification as local variables?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/models/AttributeData.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.Data;
+import uk.gov.hmcts.reform.amlib.enums.Permission;
+import uk.gov.hmcts.reform.amlib.enums.SecurityClassification;
+
+import java.util.Set;
+
+@Data
+@Builder
+@AllArgsConstructor
+public final class AttributeData {
```
geek-soft: I'd create sub package called `uk.gov.hmcts.reform.amlib.internal.models.query` to group all query objects.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/utils/SecurityClassifications.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.EnumSet;
+import java.util.Set;
+
+import static java.util.stream.Collectors.toSet;
+
+public final class SecurityClassifications {
+
+    private SecurityClassifications() {
+        //NO-OP
```
geek-soft: Shall we throw an exception, it seemed to be more recent pattern.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/utils/SecurityClassifications.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    /**
+     * Builds a set of security classifications including and below a given hierarchy.
+     *
+     * @param securityClassificationForRole integer value of security classification defined in
+     * {@link SecurityClassification}
+     * @return a set of security classifications
+     */
+    public static Set<SecurityClassification> fromValueOf(int securityClassificationForRole) {
```
geek-soft: I'd think better name. Maybe `getSecurityClassificationsAtSameOrLowerHirarchy` or `getVisibleSecurityClassifications`

Same way we should seek for better name for argument. It is not important what input security classification is or comes from. Utility method should be generic.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/RolePermissions.java
```diff
@@ -0,0 +1,20 @@
...
+import uk.gov.hmcts.reform.amlib.enums.SecurityClassification;
+
+import java.util.Map;
+import java.util.Set;
+
+@Data
+@Builder
+public final class RolePermissions {
+    private final Map<JsonPointer, Set<Permission>> permissions;
+    private final AccessType accessType;
```
geek-soft: We said during planning to name it `roleAccessType`. 

This way we will have:

```
RolePermissions:
  permissions:
  securityClassifications:
  roleSecurityClassification: 
  roleAccessType:
```

I also find above order of properties bit more tidy :)

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/RolePermissions.java
```diff
@@ -0,0 +1,20 @@
...
+    private final SecurityClassification roleSecurityClassification;
+    private final Map<JsonPointer, SecurityClassification> securityClassification;
```
geek-soft: Since it is map we should use plural so `securityClassifications` not `securityClassification`

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/internal/utils/SecurityClassificationTest.java
```diff
@@ -42,4 +54,24 @@ void whenRoleSecurityClassificationIsHighEnoughResourceShouldBeVisible(boolean e
...
+        return Stream.of(
+            new FromValueArguments(NONE, NONE),
+            new FromValueArguments(PUBLIC, NONE, PUBLIC),
+            new FromValueArguments(PRIVATE, NONE, PUBLIC, PRIVATE),
+            new FromValueArguments(RESTRICTED, NONE, PUBLIC, PRIVATE, RESTRICTED)
+        );
+    }
+
+    private static class FromValueArguments {
+        private final SecurityClassification roleSecurityClassification;
```
geek-soft: I'd be keen to avoid using word `role` here, although this is just example usage so might be acceptable.

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/internal/utils/SecurityClassificationTest.java
```diff
@@ -42,4 +54,24 @@ void whenRoleSecurityClassificationIsHighEnoughResourceShouldBeVisible(boolean e
...
+        private final SecurityClassification[] securityClassifications;
+
+        private FromValueArguments(SecurityClassification roleSecurityClassification,
+                                   SecurityClassification... securityClassifications) {
```
geek-soft: Maybe calling it `expectedSecurityClassifications` would make it bit more readable. I am generally not a big fan of such classes that holds both input and expected output but since you wanted to use varargs you had no choice.

Alternatively you could consider `Arguments.of` as below:

```
Arguments.of(NONE, ImmutableSet.of(NONE)),
Arguments.of(PUBLIC, ImmutableSet.of(NONE, PUBLIC)),
Arguments.of(PRIVATE, ImmutableSet.of(NONE, PUBLIC, PRIVATE)),
Arguments.of(RESTRICTED, ImmutableSet.of(NONE, PUBLIC, PRIVATE, RESTRICTED))
```

In my mind it is bit more readable as you segregate more what is the input and output especially that type for both is same.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
swalker125: oversight from me, left over from previous method

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
swalker125: sure

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/utils/SecurityClassifications.java
swalker125: have updated along with Permissions.class

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/internal/utils/SecurityClassificationTest.java
swalker125: I prefer the Arguments.of() way. Less code 👍

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
swalker125: access type doesn't make sense, only used once. Sec classification, sure

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
swalker125: already renamed 👍

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -263,6 +246,47 @@ public FilteredResourceEnvelope filterResource(@NotBlank String userId,
...
+     * Retrieves {@link RolePermissions} filtered by role security classification for a specific resource and role.
+     *
+     * @param resourceDefinition {@link ResourceDefinition} a unique service name, resource type and resource name
+     * @param roleName           user role name
+     * @return {@link RolePermissions} a combination of permissions and security classifications for a role name
+     * @throws PersistenceException if any persistence errors were encountered
+     */
+    @AuditLog("returned role access to resource defined as '{{resourceDefinition.serviceName}}|"
+        + "{{resourceDefinition.resourceType}}|{{resourceDefinition.resourceName}}' for role '{{roleName}}': "
+        + "{{result}}")
```
geek-soft: @swalker125 can you paste exemplar audit log? I wonder how `{{result}}` is rendered.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/repositories/AccessManagementRepository.java
```diff
@@ -49,6 +50,12 @@
     @RegisterConstructorMapper(RoleBasedAccessRecord.class)
     List<RoleBasedAccessRecord> getRolePermissions(@BindBean ResourceDefinition resourceDefinition, String roleName);
 
+    @SqlQuery("select distinct d.attribute, d.permissions, ra.default_security_classification from default_permissions_for_roles d"
+        + " join resource_attributes ra on d.service_name = ra.service_name and d.resource_type = ra.resource_type and d.resource_name = ra.resource_name and d.attribute = ra.attribute"
```
geek-soft: Minor inconsistency, sometimes we have space at the beginning of the line, sometimes at the end. No consistency so for, possible improvement for follow up PR some day.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/utils/SecurityClassifications.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    /**
+     * Builds a set of security classifications including and below a given hierarchy.
+     *
+     * @param securityClassificationForRole integer value of security classification defined in
+     * {@link SecurityClassification}
+     * @return a set of security classifications
+     */
+    public static Set<SecurityClassification> getVisibleSecurityClassifications(int securityClassificationForRole) {
```
geek-soft: Argument name change is still outstanding. As I pointed earlier mentioning role does not make much sense in generic utility class.

am-lib/src/test/java/uk/gov/hmcts/reform/amlib/internal/utils/SecurityClassificationTest.java
```diff
@@ -13,17 +15,29 @@
...
+                                                                           SecurityClassification roleClassification,
+                                                                           SecurityClassification resourceClassification) {
         boolean isVisible = resourceClassification.isVisible(roleClassification.getHierarchy());
         assertThat(isVisible).isEqualTo(expectedResult);
     }
 
-    private static Stream<Arguments> createArguments() {
+    @ParameterizedTest
+    @MethodSource("fromValueArguments")
+    void fromValueOfShouldReturnSetOfSecurityClassifications(SecurityClassification securityClassification,
```
geek-soft: Test could use rename since main method got renamed. Same with static method playing role of data source.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetRolePermissionsIntegrationTest.java
```diff
@@ -13,127 +13,171 @@
...
 import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.RESOURCE_TYPE;
 import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.ROLE_NAME;
-import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.ROLE_NAMES;
 import static uk.gov.hmcts.reform.amlib.helpers.TestConstants.SERVICE_NAME;
 
 class GetRolePermissionsIntegrationTest extends PreconfiguredIntegrationBaseTest {
     private static AccessManagementService service = initService(AccessManagementService.class);
     private static DefaultRoleSetupImportService importerService = initService(DefaultRoleSetupImportService.class);
 
+    private Map<JsonPointer, Map.Entry<Set<Permission>, SecurityClassification>> attributePermissionsForRole;
```
geek-soft: I am not sure we have a good reason for keeping this as shared class field. Technically it should be local variable declared in the body of method. I'd suggest changing it.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetRolePermissionsIntegrationTest.java
```diff
@@ -13,127 +13,171 @@
...
+    private String roleName;
+
     @BeforeEach
     void setUp() {
+        roleName = UUID.randomUUID().toString();
+    }
+
+    @Test
+    void whenGettingRolePermissionsShouldReturnPermissionsAndSecurityClassificationsForSpecifiedRole() {
         Map.Entry<Set<Permission>, SecurityClassification> readPermission = new Pair<>(ImmutableSet.of(READ), PUBLIC);
```
geek-soft: Since it is not only permission but also security classification I think we should say `publicReadPermission` and `restrictedReadPermission`. Even when we use one SC only it will still be easier to follow logic.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetRolePermissionsIntegrationTest.java
```diff
@@ -13,127 +13,171 @@
...
                 .resourceName(RESOURCE_NAME)
                 .build())
-            .attributePermissions(attributePermissionsForOtherRole)
+            .attributePermissions(attributePermissionsForRole)
             .build());
+
+        RolePermissions rolePermissions =
+            service.getRolePermissions(buildResource(SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME), roleName);
+
+        assertThat(rolePermissions.getPermissions())
```
geek-soft: I have a feeling doing `isEqual(RolePermissions.builder()...build())` might be more readable. Maybe worth trying.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetRolePermissionsIntegrationTest.java
```diff
@@ -13,127 +13,171 @@
...
+            .build());
+
+        RolePermissions rolePermissions =
+            service.getRolePermissions(buildResource(SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME), roleName);
+
+        assertThat(rolePermissions).isNull();
+    }
+
+    @Test
+    void whenRolePermissionsHasSecurityClassificationShouldRemoveAttributesWithHigherSecurityClassification() {
```
geek-soft: A `whenRolePermissionsHasSecurityClassification` part of method name is bit ambiguous. Maybe `whenGettingRolePermissionsShouldRemoveAttributesWithHigherSecurityClassification` be enough.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetRolePermissionsIntegrationTest.java
```diff
@@ -13,127 +13,171 @@
...
+        assertThat(rolePermissions.getRoleAccessType()).isEqualTo(ROLE_BASED);
+
+        assertThat(rolePermissions.getRoleSecurityClassification()).isEqualTo(PUBLIC);
     }
 
     @Test
-    void returnListOfPermissionsForRoleName() {
-        Map<JsonPointer, Set<Permission>> accessRecord =
-            service.getRolePermissions(buildResource(SERVICE_NAME, RESOURCE_TYPE, RESOURCE_NAME), ROLE_NAMES);
+    void whenInsufficientRolePermissionsShouldReturnNull() {
```
geek-soft: I'd move it one method down.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/RolePermissions.java
```diff
@@ -0,0 +1,20 @@
...
+import uk.gov.hmcts.reform.amlib.enums.Permission;
+import uk.gov.hmcts.reform.amlib.enums.SecurityClassification;
+
+import java.util.Map;
+import java.util.Set;
+
+@Data
+@Builder
+public final class RolePermissions {
+    private final Map<JsonPointer, Set<Permission>> permissions;
```
abhijitdiwan: As per requirement we are returning this but we are not returning relationships (Claimant or defendant)

swalker125: relationships is only relevant to explicit access. This deals with role based access. Am I missing something?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
swalker125: 2019-04-29T06:55:50.354 INFO  [main] u.g.h.reform.amlib.internal.aspects.AuditingAspect [Access Management audit]: returned role access to resource defined as 'CMC|Case|Claim' for role 'ec02af28-982d-4e1a-a4e8-f460e2ac7615': RolePermissions(permissions={/child=[READ], /parent/age=[READ]}, securityClassifications={/child=PUBLIC, /parent/age=PUBLIC}, roleSecurityClassification=PUBLIC, roleAccessType=ROLE_BASED)

swalker125: I guess this would be better split as result.permissions etc?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/repositories/AccessManagementRepository.java
swalker125: sure let's leave for a clean up PR. Should be quick and do-able before Tuesday

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/utils/SecurityClassifications.java
swalker125: changed to hierarchy

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetRolePermissionsIntegrationTest.java
swalker125: sure, it was just to avoid repetition of Map<.....>. changed

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
swalker125: maybe only result.permissions is important actually... Shall discuss this morning when we're in the office.

geek-soft: I think it would be better to have custom format instead of relying on RolePermissions#toString. Please propose something.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/models/RolePermissions.java
geek-soft: Nope, you are right @swalker125

PR #144, opened 2019-04-25
build: add checksum check to nightly pipeline

rebeccabrennan:
Checksum added to Jenkins nightly file to improve security.


test.sh
```diff
@@ -0,0 +1,7 @@
+set -e
```
geek-soft: What is that needed for?

Jenkinsfile_nightly
```diff
@@ -42,8 +42,24 @@ withNightlyPipeline(type, product, component) {
...
-    sh 'PATH=$PATH:./bin git lfs install --local && PATH=$PATH:./bin git lfs pull'
+      sh '''
+        set -e
+    
+        echo "7be85238cbbb957ab25de52b60279d40ba40d3faa72eeb2cb9fa77d6d92381e5  git-lfs-v2.7.1.tar.gz" > git-lfs-v2.7.1.checksum.txt
+        curl -L https://github.com/git-lfs/git-lfs/releases/download/v2.7.1/git-lfs-linux-amd64-v2.7.1.tar.gz > git-lfs-v2.7.1.tar.gz
+        sha256sum -c git-lfs-v2.7.1.checksum.txt
+        
+        tar xz -C bin/ git-lfs -f git-lfs-v2.7.1.tar.gz
+        ls -lash bin/
```
geek-soft: I guess that listing is not really needed.

Jenkinsfile_nightly
```diff
@@ -42,8 +42,24 @@ withNightlyPipeline(type, product, component) {
...
+      '''
+
+      sh '''
+        set -e
+      
+        export PATH=$PATH:./bin
+        git lfs install --local
+        git lfs pull || (git lfs logs last && false)
```
geek-soft: Please explain what `git lfs logs last && false` does and why is that needed.

Jenkinsfile_nightly
```diff
@@ -42,8 +42,24 @@ withNightlyPipeline(type, product, component) {
...
+    
+        echo "7be85238cbbb957ab25de52b60279d40ba40d3faa72eeb2cb9fa77d6d92381e5  git-lfs-v2.7.1.tar.gz" > git-lfs-v2.7.1.checksum.txt
+        curl -L https://github.com/git-lfs/git-lfs/releases/download/v2.7.1/git-lfs-linux-amd64-v2.7.1.tar.gz > git-lfs-v2.7.1.tar.gz
+        sha256sum -c git-lfs-v2.7.1.checksum.txt
+        
+        tar xz -C bin/ git-lfs -f git-lfs-v2.7.1.tar.gz
+        ls -lash bin/
+      '''
+
+      sh '''
```
geek-soft: I do not see a reason to have two bash scripts. I suggest combining these together as they are both aim to achieve common goal.

Jenkinsfile_nightly
```diff
@@ -42,8 +42,24 @@ withNightlyPipeline(type, product, component) {
...
 
   after('checkout') {
-    sh 'curl -L https://github.com/git-lfs/git-lfs/releases/download/v2.7.1/git-lfs-linux-amd64-v2.7.1.tar.gz | tar xz -C ./bin git-lfs'
-    sh 'PATH=$PATH:./bin git lfs install --local && PATH=$PATH:./bin git lfs pull'
+      sh '''
+        set -e
+    
+        echo "7be85238cbbb957ab25de52b60279d40ba40d3faa72eeb2cb9fa77d6d92381e5  git-lfs-v2.7.1.tar.gz" > git-lfs-v2.7.1.checksum.txt
+        curl -L https://github.com/git-lfs/git-lfs/releases/download/v2.7.1/git-lfs-linux-amd64-v2.7.1.tar.gz > git-lfs-v2.7.1.tar.gz
+        sha256sum -c git-lfs-v2.7.1.checksum.txt
```
geek-soft: Can you suppress output? There might be `--status` flag to return status code only.

Jenkinsfile_nightly
```diff
@@ -42,8 +42,24 @@ withNightlyPipeline(type, product, component) {
...
+        
+        tar xz -C bin/ git-lfs -f git-lfs-v2.7.1.tar.gz
```
geek-soft: Please change order of arguments to make it more readable. I suggest:

```
tar xz -f git-lfs-v2.7.1.tar.gz -C bin/ git-lfs
```

This way it naturally says input file, output dir and finally output file.

Jenkinsfile_nightly
rebeccabrennan: if git lfs fails to startup it'll get the logs and then fail the build

geek-soft: So how does it help? If we bring some code we should know what it does, am I right?

geek-soft: Also I am interested to know what was the initial issue with git-lfs. Can you explain me why it used to fail?

Jenkinsfile_nightly
rebeccabrennan: 👍

Jenkinsfile_nightly
rebeccabrennan: 👍

Jenkinsfile_nightly
rebeccabrennan: it did not like that it was in the build folder removing it from there made it work

rebeccabrennan: 👍

Jenkinsfile_nightly
rebeccabrennan: 👍

test.sh
rebeccabrennan: set e means fail on errors. I was taking security precautions I felt may be helpful. Although removing the test.sh script as we do not need it anymore do we?

Jenkinsfile_nightly
rebeccabrennan: we could try something like https://unix.stackexchange.com/questions/119648/redirecting-to-dev-null/119650#119650 do you think this would be helpful though there is only minimal output

test.sh
geek-soft: I meant whole file.

Jenkinsfile_nightly
geek-soft: Have you tried `--status` flag?

test.sh
rebeccabrennan: yeah removed

Jenkinsfile_nightly
rebeccabrennan: where do you want the --status flag ?

PR #197, opened 2019-05-29
feat: CCD to AM migration scripts (AM-266)

bmcts:
https://tools.hmcts.net/jira/browse/AM-266

### Change description ###
Added scripts to migrate CCD data to AM database. One script for initial population of supporting AM tables (to be removed later as CCD will populate these fields themselves) and another script for the main data migration to the access_management table.

### DIFF
--- a/migration/ccd-to-am-migration-main.sql
+++ b/migration/ccd-to-am-migration-main.sql
@@ -8,7 +8,6 @@ CREATE TEMP TABLE stage
 )
 ON COMMIT DROP;
 
-ALTER TABLE access_management DROP CONSTRAINT "AccessManagement_pkey";
 ALTER TABLE access_management DROP CONSTRAINT access_management_unique;
 ALTER TABLE access_management DROP CONSTRAINT access_management_resources_fkey;
 ALTER TABLE access_management DROP CONSTRAINT relationship_fkey;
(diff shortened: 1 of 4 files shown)


migration/ccd-to-am-migration-runner.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+
+echo
+set -e
+set -u
+
+DB="am"
+USER="amuser"
```
shashank-rastogi: DB hostname/IP is also required to connect to a remote db

migration/ccd-to-am-migration-main.sql
```diff
@@ -0,0 +1,57 @@
...
+        AND re.resource_name = s.case_type_id
+        AND s.case_role IN (SELECT role_name FROM roles)
+    EXCEPT
+        SELECT resource_id, accessor_type, accessor_id, "attribute",
+            permissions, service_name, resource_name, resource_type,
+            relationship
+        FROM access_management
+    RETURNING resource_id
+)
+SELECT COUNT(*) AS "access_management inserts" FROM ins_access_management;
```
shashank-rastogi: i would suggest to run any queries for data validation, metrics collection after the commit

shashank-rastogi (changes requested):

migration/ccd-to-am-migration-runner.sh
```diff
@@ -0,0 +1,31 @@
...
+INIT_SQL="./ccd-to-am-migration-init.sql"
```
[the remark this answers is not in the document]
bmcts: psql doesn't have a password input option so it always has to be entered through an interactive prompt

migration/ccd-to-am-migration-main.sql
bmcts: it's difficult to separate the data insertion and metric collection in this case as they're both part of the same SQL query... should they be separated into two different queries? It would be less efficient but we could put a commit between the two queries

migration/ccd-to-am-migration-runner.sh
bmcts: 👍

shashank-rastogi (approved):

PR #227, opened 2019-06-26
feat: optional security classification methods (AM-323)

bmcts:
https://tools.hmcts.net/jira/browse/AM-323

### DIFF
--- a/am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
+++ b/am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
@@ -303,10 +303,11 @@ public class AccessManagementService {
         List<ExplicitAccessRecord> explicitAccessRecords) {
 
         List<Map<JsonPointer, Set<Permission>>> permissionsForRelationships = explicitAccessRecords.stream()
-                .collect(collectingAndThen(groupingByWithNullKeys(ExplicitAccessRecord::getRelationship), Map::values)).stream()
-                    .map(explicitAccessRecords -> explicitAccessRecords.stream()
-                        .collect(getMapCollector()))
-                .collect(toList());
+            .collect(collectingAndThen(groupingByWithNullKeys(ExplicitAccessRecord::getRelationship), Map::values))
+            .stream()
+                .map(explicitAccessRecordsForRelationship -> explicitAccessRecordsForRelationship.stream()
+                    .collect(getMapCollector()))
+            .collect(toList());
 
         return permissionsService.merge(permissionsForRelationships);
     }
--- a/am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
+++ b/am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
@@ -39,8 +39,6 @@ import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotEmpty;
 import javax.validation.constraints.NotNull;
 
-import static java.util.Objects.isNull;
-import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.collectingAndThen;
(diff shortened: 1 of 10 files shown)


am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
```diff
@@ -154,8 +172,29 @@ public void revokeResourceAccess(@NotNull @Valid ExplicitAccessMetadata accessMe
...
 
     /**
-     * Filters {@link JsonNode} to remove fields that user has no access to (no READ permission). In addition to that
-     * method also returns map of all permissions that user has to resource.
+     * Filters {@link JsonNode} to remove fields that user has no access to (no READ permission or insufficient
+     * security classification). In addition to that method also returns map of all permissions that user has to
+     * resource.
+     *
+     * @param userId                           accessor ID
+     * @param userRoles                        accessor roles
```
shashank-rastogi: not sure if we need to override with so many filterResource() methods. i think follwoing should suffice -

//parent method for the filter resource w/o security classification - this should be exposed as a new API method
 public List<FilteredResourceEnvelope> filterResource(String userId, Set userRoles, List resources) {
             return resources.stream()
            .map(resource -> filterResource(userId, userRoles, resource, null))
            .collect(toList());
}

//parent method for the filter resource with security classification - this should be exposed as a new API method
public List<FilteredResourceEnvelope> filterResource(String userId, Set userRoles, Listresources, SecurityClassification attributeSecurityClassifications) {

 //inThrows NoSuchElementException exception when root is missing
        if (isNull(attributeSecurityClassifications.get(JsonPointer.valueOf("")))) {
            throw new NoSuchElementException("Root element not found in input Security Classification");
        }
        return resources.stream()
            .map(resource -> filterResource(userId, userRoles, resource, attributeSecurityClassifications))
            .collect(toList());
}

//actual implementation method for both flows
public FilteredResourceEnvelope filterResource(String userId,
                                                   Set userRoles,
                                                   List resource,  SecurityClassification  attributeSecurityClassifications) {
.....
}

shashank-rastogi (changes requested):

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/AccessManagementService.java
bmcts: Agree that these methods need to be cleaned up, but let's just clarify what each method actually does to see whether they are required or not.

The filterResource methods that take a list of resources as input are not currently supported by the API. To support them would require new endpoints, new input classes to be created (which expect lists of resources rather than just a single resource), and all of the tests to be updated to reflect this. A lot of rework required and would not be completed by the end of the sprint.

The filterResource method that takes only 3 parameters (without security classification) is not particularly required for the API either (but is currently used). This is because when the API method is called without security classification, the serialisation just sets the attribute security classification as null. The main filterResource method (where the main filtering logic is) actually expects the attribute security classification to be null if filtering by security classification is not required, so when null is passed as input it will run as expected. The only reason to have the filterResource method with 3 parameters is for the library to have a filterResource method that does not require attribute security classification to be explicitly stated, and then setting it to null afterwards, essentially abstracting away the concept of security classification from the consumer of the library. The way things are going, it doesn't seem like anyone will be consuming the library (just using the API instead), so it is not clear whether the filterResource method with 3 parameters will be useful in the long run.

The filterResource method that takes 4 parameters (including security classification), is required for the API and is currently used. It can handle either null or not null security classification for a single resource, and simply passes the information on to the main filtering logic. Definitely required.

My thoughts: firstly the filterResource methods that take lists of resources should be renamed to filterResources. API support for these methods should be put in a new story and done in a future sprint. For this sprint, I don't think any change is required. Having 2 overloaded filterResource methods is not excessive, and it keeps separation of the methods for the sake of the library.

Let me know what you think

shashank-rastogi (approved):

PR #268, opened 2019-07-26
fix: Gatling test initial data load

abhijitdiwan:
https://tools.hmcts.net/jira/browse/AM-362
https://tools.hmcts.net/jira/browse/AM-380

### Change description ###
fix: Gatling test initial data load

### DIFF
--- a/Jenkinsfile_nightly
+++ b/Jenkinsfile_nightly
@@ -88,10 +88,9 @@ withNightlyPipeline(type, product, component) {
                         "load" \
                         ${DATABASE_USER} \
                         ${DATABASE_NAME} \
+                        ${DATABASE_HOST} \
                         ${DATABASE_PASS} \
-                        ${DATABASE_HOST} \
                         ${DATABASE_PORT} \
-                        ${DATABASE_HOST} \
                         ${params.ENVIRONMENT}"
         }
       }
--- a/Jenkinsfile_nightly
+++ b/Jenkinsfile_nightly
@@ -17,15 +17,6 @@ def type = "java"
 def product = "am"
 def component = "lib-test"
 
-
-/*List<Map<String, Object>> secrets = [
-  secret('am-lib-test-POSTGRES-HOST', 'DATABASE_HOST'),
-  secret('am-lib-test-POSTGRES-PORT', 'DATABASE_PORT'),
-  secret('am-lib-test-POSTGRES-DATABASE', 'DATABASE_NAME'),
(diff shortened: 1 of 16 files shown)


src/gatling/simulations/uk/gov/hmcts/reform/amlib/performance/scenarios/BasicScenarios.scala
```diff
@@ -34,7 +34,7 @@ object BasicScenarios {
 
   val returnResourceAccessors: ScenarioBuilder = scenario("Return Resource Accessors")
     .forever(
-      feed(ReturnResourceAccessors.feed)
+      feed(CreateResourceAccess.feed)
         .exec(AccessManagement.createResourceAssess)
```
shashank-rastogi: i  am not sure why this change is required?

Jenkinsfile_nightly
```diff
@@ -59,9 +68,45 @@ withNightlyPipeline(type, product, component) {
...
+          sh "/nightlyperformancedata/load-or-delete-performance-testdata.sh \
+                        ${WORKSPACE} \
+                        'delete' \
+                        ${DATABASE_USER} \
+                        ${DATABASE_NAME} \
+                        ${DATABASE_HOST} \
+                        ${DATABASE_PASS} \
+                        ${DATABASE_PORT} \
+                        ${params.ENVIRONMENT}"
+        }
```
shashank-rastogi: as this could be used to set up data for other types of tests like functional tests as well, we should try to create a function and pass parameter like test type and script to be executed and try to avoid writing same code multiple times

shashank-rastogi (changes requested):

src/gatling/simulations/uk/gov/hmcts/reform/amlib/performance/scenarios/BasicScenarios.scala
abhijitdiwan: This is just scala class setting default values for feed def feed: Feeder[String] = Iterator.continually {
    Map(
      "resourceId" -> randomUUID,
      "accessorId" -> randomUUID
    )
  }

we don't have any ReturnResourceAccessors feed.  and we can use the same.

Jenkinsfile_nightly
abhijitdiwan: done

shashank-rastogi (approved):

PR #293, opened 2019-08-13
test: serenity test data load

abhijitdiwan:
https://tools.hmcts.net/jira/browse/AM-338

### Change description ###
test: serenity test data load


bmcts (approved):
Approved with the condition that the tech debt created here will be revisited in a future sprint. The functional tests use the API in a higher environment (AAT), which means that it runs the tests against the previous build of the API, not the current build. This will cause problems when introducing functional tests for new endpoints being added to the API; the endpoints won't exist in the previous build causing the new tests to fail.

PR #325, opened 2019-08-29
feat: aks migration readiness

bmcts:


charts/am-api/values.yaml
```diff
@@ -7,13 +7,5 @@ am-lib-test:
     AM_DB_PORT: 5432
     AM_DB_USERNAME: "{{ .Values.postgresql.postgresqlUsername }}"
     ENABLE_DB_MIGRATE_IN_SERVICE: true
-  image: "https://hmcts.azurecr.io/hmcts/am-lib-test:latest"
-  keyVaults:
-    "s2s":
```
adusumillipraveen: How are you getting this secret now ?

bmcts: it is being set as an environment variable in the Jenkinsfile

adusumillipraveen: No, that won't work.. you need to use keyvaults as flex volumes .

PR #337, opened 2019-09-03
feat: Versioning Api changes AM-381

abhijitdiwan:
https://tools.hmcts.net/jira/browse/AM-381

### Change description ###
feat: Versioning Api changes AM-381


am-lib/src/main/resources/db/migration/migrationAks/V19__load_data.sql
```diff
@@ -0,0 +1,9 @@
+insert into services (service_name, service_description) values ('cmc-test', null)
+on conflict on constraint services_pkey do nothing;
+
+insert into resources (service_name, resource_type, resource_name) values ('cmc-test', 'case-test', 'claim-test')
+ on conflict on constraint resources_pkey do nothing;
+
+insert into roles (role_name, role_type, security_classification, access_type) values ('caseworker-test', cast('RESOURCE' as role_type),cast('PUBLIC' as security_classification), cast('ROLE_BASED' as access_type))
```
shashank-rastogi: why this is required for versioning?

I believe these are the changes to fix the issue with running functional tests in local. If yes, then please move them to separate PR as they should not be part of the versioning feature

src/functionalTest/java/uk/gov/hmcts/reform/amapi/functional/AccessManangementApiTest.java
```diff
@@ -29,6 +29,7 @@
 
     Response response;
 
+    public static final String VERSION = "v1";
```
shashank-rastogi: instead of defining version as static variable in the class, can we read this from application-functional.yaml?

PR #345, opened 2019-09-04
feat: Versioning Api changes AM-381

abhijitdiwan:
https://tools.hmcts.net/jira/browse/AM-381

### Change description ###
feat: Versioning Api changes AM-381


src/functionalTest/java/uk/gov/hmcts/reform/amapi/functional/AccessManangementApiTest.java
```diff
@@ -29,6 +29,7 @@
 
     Response response;
 
+    public static final String VERSION = "v1";
```
shashank-rastogi: instead of defining version as static variable in the class, can we read this from application-functional.yaml?

abhijitdiwan: updated. We can used from yaml for Function and controller and Integration tests. For unit and Gatling we can't use.

PR #366, opened 2019-09-11
feat: AM add auditing functionality (AM-376)

abhijitdiwan:
https://tools.hmcts.net/jira/browse/AM-376

### Change description ###
feat: AM add auditing functionality (AM-376)


am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/models/ExplicitAccessRecord.java
```diff
@@ -24,6 +26,8 @@
     private final JsonPointer attribute;
     private final Set<Permission> permissions;
     private final String relationship;
+    @Nested
+    private final AccessManagementAudit accessManagementAudit;
```
bmcts: Do we really need to put the auditing information into its own AccessManagementAudit object? Having two fields, lastUpdate and callingServiceName, would make things much easier - no need for @Nested annotations or custom constructors, and would make the db queries less complex

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/models/Role.java
```diff
@@ -15,4 +17,5 @@
     private final RoleType roleType;
     private final SecurityClassification securityClassification;
     private final AccessType accessType;
+    private LocalDateTime lastUpdate;
```
bmcts: I'm not sure using LocalDateTime for timestamps is suitable for auditing. LocalDateTime omits the timezone, which means that it doesn't capture the actual moment that the transaction happened. It would be better to use java.time.Instant which captures the moment in UTC

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/repositories/DefaultRoleSetupRepository.java
```diff
@@ -15,25 +15,29 @@
...
 
     @SqlUpdate("insert into roles (role_name, role_type, security_classification, access_type) values (:roleName, cast(:roleType as role_type), cast(:securityClassification as security_classification), cast(:accessType as access_type))"
-        + " on conflict on constraint roles_pkey do update set role_type = cast(:roleType as role_type), security_classification = cast(:securityClassification as security_classification), access_type = cast(:accessType as access_type)")
+        + " on conflict on constraint roles_pkey do update set role_type = cast(:roleType as role_type), security_classification = cast(:securityClassification as security_classification), access_type = cast(:accessType as access_type), last_update = now()")
     void addRole(String roleName, RoleType roleType, SecurityClassification securityClassification, AccessType accessType);
 
     @SqlUpdate("insert into resources (service_name, resource_type, resource_name) values (:serviceName, :resourceType, :resourceName)"
-        + "on conflict on constraint resources_pkey do nothing")
+        + "on conflict on constraint resources_pkey do update set last_update = now()")
```
bmcts: Some of the queries use CURRENT_TIMESTAMP, some use now(), can we make it consistent?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/models/ExplicitAccessRecord.java
abhijitdiwan: i think it should reused across objects. it is good to use object.If we want to extend it it would be easy. else adding every property in every object would be redundant.

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/models/Role.java
abhijitdiwan: We can use instant..But I think we are just using one time zone.out of one timezone it won't get used right?

am-lib/src/main/java/uk/gov/hmcts/reform/amlib/internal/repositories/DefaultRoleSetupRepository.java
abhijitdiwan: Yes correct will update it.

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/importer/DefaultPermissionIntegrationTest.java
```diff
@@ -82,6 +85,39 @@ void shouldOverwriteExistingRecordWhenEntryIsAddedASecondTime() {
...
+        final Instant localDateTime = roleBasedAccessRecord.getAccessManagementAudit().getLastUpdate();
+        assertThat(roleBasedAccessRecord).isNotNull();
+        assertThat(roleBasedAccessRecord.getAccessManagementAudit().getCallingServiceName()).isNotNull();
+        assertThat(roleBasedAccessRecord.getAccessManagementAudit().getCallingServiceName())
+            .isEqualTo("integration-test");
+        assertThat(localDateTime).isNotNull();
+
+        //Update Audit
+        audit = AccessManagementAudit.builder().lastUpdate(Instant.now())
+            .callingServiceName("integration-test123").build();
```
shashank-rastogi: can we keep service name as meaningful?

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/GetRolePermissionsIntegrationTest.java
```diff
@@ -155,17 +163,73 @@ void shouldReturnNullWhenDefaultRoleNameDoesNotExist() {
...
+    @Test
+    void whenAuditLogsForRolePermissionAttributesThenShouldReturnsAuditDetails() {
+
+        Map.Entry<Set<Permission>, SecurityClassification> publicReadPermission =
+            new Pair<>(ImmutableSet.of(READ), PUBLIC);
+
+
+        //Add Audit
+        AccessManagementAudit audit = AccessManagementAudit.builder().lastUpdate(Instant.now())
+            .callingServiceName("Integration-test").build();
```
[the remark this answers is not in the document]
abhijitdiwan: These are Java unit case not spring ones. adding yaml will be more overhead. and these are mock values

am-lib/src/test/java/integration/uk/gov/hmcts/reform/amlib/importer/DefaultPermissionIntegrationTest.java
abhijitdiwan: renamed.

